PR #24, opened 2020-11-07
More FreeBSD fixes

ZmnSCPxj:
@hosiawak please check: 
[clboss-more-freebsd.tar.gz](https://github.com/ZmnSCPxj/clboss/files/5505187/clboss-more-freebsd.tar.gz)

This should no longer require an explicit `CXXFLAGS="-Wno-noexcept-type"`, it "should" suppress the warning on FreeBSD even with a plain `./configure`.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -430,7 +430,7 @@ EXTRA_DIST = \
 	external/jsmn/LICENSE \
 	external/jsmn/jsmn.h
 
-AM_CXXFLAGS = -Wall -Werror $(PTHREAD_CFLAGS) $(libev_CFLAGS) $(SQLITE3_CFLAGS) $(CURL_CFLAGS)
+AM_CXXFLAGS = -Wall -Werror $(PTHREAD_CFLAGS) $(libev_CFLAGS) $(SQLITE3_CFLAGS) $(CURL_CFLAGS) $(CLBOSS_CXXFLAGS)
 LDADD = libclboss.la $(PTHREAD_LIBS) $(libev_LIBS) $(SQLITE3_LIBS) $(CURL_LIBS)
 
 ACLOCAL_AMFLAGS = -I m4
--- a/Net/ProxyConnector.cpp
+++ b/Net/ProxyConnector.cpp
@@ -1,5 +1,6 @@
 #include<errno.h>
 #include<netdb.h>
+#include<netinet/in.h>
 #include<string.h>
 #include<sys/socket.h>
 #include<sys/types.h>
--- a/Net/SocketFd.cpp
+++ b/Net/SocketFd.cpp
@@ -1,5 +1,6 @@
 #include<errno.h>
 #include<stdexcept>
+#include<string>
 #include<string.h>
 #include<sys/socket.h>
 #include<unistd.h>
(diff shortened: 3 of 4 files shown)


hosiawak:
Without the explicit `CXXFLAGS="-Wno-noexcept-type"` it gave me this:

```
In file included from Boss/Mod/Rpc.cpp:13:
/usr/local/include/ev.h:548:18: error: mangled name of 'ev_set_allocator' will change in C++17 due to non-throwing exception specification in function signature
      [-Werror,-Wc++17-compat-mangling]
EV_API_DECL void ev_set_allocator (void *(*cb)(void *ptr, long size) EV_NOEXCEPT) EV_NOEXCEPT;
                 ^
/usr/local/include/ev.h:554:18: error: mangled name of 'ev_set_syserr_cb' will change in C++17 due to non-throwing exception specification in function signature
      [-Werror,-Wc++17-compat-mangling]
EV_API_DECL void ev_set_syserr_cb (void (*cb)(const char *msg) EV_NOEXCEPT) EV_NOEXCEPT;
                 ^
/usr/local/include/ev.h:675:18: error: mangled name of 'ev_set_loop_release_cb' will change in C++17 due to non-throwing exception specification in function signature
      [-Werror,-Wc++17-compat-mangling]
EV_API_DECL void ev_set_loop_release_cb (EV_P_ void (*release)(EV_P) EV_NOEXCEPT, void (*acquire)(EV_P) EV_NOEXCEPT) EV_NOEXCEPT;
```

The good news is with the flag and with `--disable-dependency-tracking` passed to configure it compiles using `gmake` and runs 👍 :

```
# clboss --version
clboss 0.5A
```

ZmnSCPxj:
After configuring *without* `-Wno-noexcept-type` can you please `grep 'Wno-noexcept-type' config.log` and `grep 'CLBOSS_CXXFLAGS' config.log` and `grep 'Wno-noexcept-type' Makefile`? Thank you.

hosiawak:
```
# grep 'Wno-noexcept-type' config.log
configure:17964: c++ -std=gnu++11 -c -g -O2 -Wno-noexcept-type  conftest.cpp >&5
```

```
# grep 'CLBOSS_CXXFLAGS' config.log
CLBOSS_CXXFLAGS=''
```

```
# grep 'Wno-noexcept-type' Makefile
```

ZmnSCPxj:
~~Thank you, can you send me the `config.log` file then?~~ Never mind, I think I figured out why the `configure` is not detecting it, will update this.

ZmnSCPxj:
Hi @hosiawak, can you try this? 
[clboss-more-freebsd.tar.gz](https://github.com/ZmnSCPxj/clboss/files/5506641/clboss-more-freebsd.tar.gz)

Should not need the `CXXFLAGS=-Wno-noexcept-type`, though probably still needs the `--disable-dependency-tracking`.  During `configure` you should see something like:

    checking if compiler compiles ev.h with -Wno-noexcept-type... yes

hosiawak:
Yes, I saw that message and it compiled without the `CXXFLAGS=-Wno-noexcept-type`. 👍

Issue #25, opened 2020-11-08
clboss-status reporting internet offline


hosiawak:
My fresh install of FreeBSD clboss-status reports internet connection being constantly offline.

I checked `Boss/Mod/InternetConnectionMonitor.cpp` and it seems to be checking some popular servers on port 443 periodically. I can connect to these servers manually.

Here's my` lightningd-bitcoin.conf`:

```
alias=[REDACTED]
lightning-dir=/var/db/c-lightning
always-use-proxy=true
large-channels
funding-confirms=1
announce-addr=[redacted].onion:1234
bind-addr=127.0.0.1:5678
bitcoin-rpcconnect=127.0.0.1
bitcoin-rpcpassword=[redacted]
bitcoin-rpcport=8332
bitcoin-rpcuser=[redacted]
log-file=lightningd.log
network=bitcoin
proxy=127.0.0.1:9050
plugin=clboss
```

Unfortunately I don't know C++ and `libev` enough to investigate this issue further but if you could maybe provide a small snippet of code I can run on this box that might help. Thank you.

ZmnSCPxj:
With a proxy, a good amount of code in the clboss gets involved (pulls in code in `Net/` directory), so a small snippet of code would not be so small.  Maybe I can add a dev binary that performs the connection check directly though.

hosiawak:
I commented out the tor proxy line and the tor announce addr and the status is now reporting online.

hosiawak:
I also noticed that when clboss was running on tor proxy (using the config above) it didn't open or rebalance any channels. It went through of `ChannelFinderByPopularity/ChannelFinderByDistance: Dijkstra progress and XXX-channeld-chan#46: Peer connection lost/Peer transient failure in CHANNELD_NORMAL: channeld: Owning subdaemon channeld died` over and over but didn't manage to do any actual work.

It looks like a possible bug with the tor proxy. I still have the logs if you need them.

hosiawak:
Running for a few hours with the proxy turned off and low fees. This thing is incredible to say the least. Opened ~ 20 channels, most of them are already relatively balanced, rebalancing doesn't seem to cost an arm and a leg either. Maybe I wasn't using the right tools on LND but before I never managed to maintain a node with the same ease, speed and low cost (like a boss ) :) Thank you 🥇

ZmnSCPxj:
[...]

PR #33, opened 2020-11-15
Net/ProxyConnector.cpp: Retry up to 100 times if the proxy reports general SOCKS5 failure.

ZmnSCPxj:
Good morning @hosiawak, I tried this workaround where we try up to 100 times to `CONNECT` if the proxy sends back a general failure, could you please check on the system where `proxy`+`always-use-proxy` leads to CLBOSS thinking we are offline?

After compiling but before restarting your node, please try this command a dozen times:

    ./dev-proxy-connect 127.0.0.1 9050 www.google.com 443

If that consistently says "Not connected" then do not bother restarting your node and just tell me.  However if it consistently says "Connected" could you please try restarting your node with this version of CLBOSS and with `proxy` + `always-use-proxy=true`?

[clboss-proxy-100.tar.gz](https://github.com/ZmnSCPxj/clboss/files/5541748/clboss-proxy-100.tar.gz)

### DIFF
--- a/Net/ProxyConnector.cpp
+++ b/Net/ProxyConnector.cpp
@@ -102,7 +102,14 @@ int to_errno(std::uint8_t socks5_err) {
 namespace Net {
 
 Net::SocketFd
-ProxyConnector::connect(std::string const& host, int port) {
+ProxyConnector::single_connect( std::string const& host, int port
+			      , bool& socks5_general_error
+			      ) {
+	/* Tor likes to send a "general proxy failure" when it is unable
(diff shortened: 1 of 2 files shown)


hosiawak:
@ZmnSCPxj It consistently says: `Not connected: Network is down (50)`

ZmnSCPxj:
Aww man... thanks for checking.  On my system it improves the success rate of that command, but it seems it does not work on yours. Hmmm.

Can you check `--socks5-hostname` with `curl`? https://github.com/ZmnSCPxj/clboss/issues/25#issuecomment-727290881

PR #34, opened 2020-11-15
Net/ProxyConnector.cpp: Send the entire `CONNECT` sequence in a single `write` call instead of multiple ones.

ZmnSCPxj:
Hi @hosiawak, can you please check this branch?
[clboss-single-proxy-write.tar.gz](https://github.com/ZmnSCPxj/clboss/files/5542316/clboss-single-proxy-write.tar.gz)

Same instructions as for #33:

After compiling but before restarting your node, please try this command a dozen times:

    ./dev-proxy-connect 127.0.0.1 9050 www.google.com 443

If that consistently says "Not connected" then do not bother restarting your node and just tell me.  However if it consistently says "Connected" could you please try restarting your node with this version of CLBOSS and with `proxy` + `always-use-proxy=true`?

Next release name if this gets fixed: Facepalm Of Doom

### DIFF
--- a/Net/ProxyConnector.cpp
+++ b/Net/ProxyConnector.cpp
@@ -145,7 +145,7 @@ ProxyConnector::connect(std::string const& host, int port) {
 	}
 
 	/* Request.  */
-	buffer.resize(4);
+	buffer.resize(4 + hostaddr.size() + 2);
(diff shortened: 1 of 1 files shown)


hosiawak:
@ZmnSCPxj `dev-proxy-connect` succeeds consistently. `clboss-status` says `online`.

Approve

PR #37, opened 2020-11-18
Json/Out.hpp: Use basic types in templates.

### LINKED ISSUE #36: `Json::Detail::Serializer` template handling of integral types
From the MacOS of @hosiawak: https://gist.github.com/hosiawak/94215c9f02d3174034b83692703b94f4

Sigh.  It is the incoherent types of C that bite us here.

A simple way would be to just use explicit basic-C `char`, `short`, `int`, `long` and `long long` types.  C++11 standard "should" support `long long` outright.  Hopefully these all cover the `std::int*_t` types on all systems.  Sigh.

Alternately, we can add more template metaprogramming and use `std::is_integral` to detect if a type is integral.  That requires us to add another template.

ZmnSCPxj:
Fixes #36 

We hope.

@hosiawak, please try on your MacOS? 
[clboss-template-basic.tar.gz](https://github.com/ZmnSCPxj/clboss/files/5558533/clboss-template-basic.tar.gz)

### DIFF
--- a/Json/Out.hpp
+++ b/Json/Out.hpp
@@ -41,69 +41,87 @@ struct Serializer<float> {
 	}
 };
 template<>
-struct Serializer<std::int8_t> {
-	static std::string serialize(std::int8_t v) {
+struct Serializer<signed char> {
+	static std::string serialize(signed char v) {
 		auto os = std::ostringstream();
 		os << std::dec << std::int64_t(v);
 		return os.str();
 	}
 };
 template<>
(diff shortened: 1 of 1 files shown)


hosiawak:
[New Sqlite error](https://gist.github.com/hosiawak/6efe655ef2d760cb849c6b91819e7a44)

ZmnSCPxj:
Looks like something vastly similar except in a different template.  Okay, will go fix.

PR #39, opened 2020-11-18
Sqlite3/Detail/*: Use basic types for templates.

ZmnSCPxj:
Hi @hosiawak, could you please check on your MacOS? [clboss-basic-template-2.tar.gz](https://github.com/ZmnSCPxj/clboss/files/5560602/clboss-basic-template-2.tar.gz)

Also can you check that these still compile on your FreeBSD?

### DIFF
--- a/Sqlite3/Detail/binds.hpp
+++ b/Sqlite3/Detail/binds.hpp
@@ -29,50 +29,62 @@ struct Bind<double> {
 };
 
 template<>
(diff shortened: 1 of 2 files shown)


hosiawak:
@ZmnSCPxj  It compiles on both systems.

ZmnSCPxj:
Wheee!

PR #43, opened 2020-11-21
--clboss-min-onchain

### LINKED ISSUE #31: Configuration - clboss-min-onchain
Edited by ZmnSCPxj

Add a `--clboss-min-onchain` option that overrides our default 0.0003 minimum amount we leave onchain.

------

Original OP:

@ZmnSCPxj Are you planning to include some way to configure clboss (eg. by passing clboss-whatever options to lightningd.conf) ?

For example `clboss-max-channels = 30` meaning it'll try to open up to that many channels. Once it reaches this threshold it'll stop the opening process (even if there are onchain funds) but it'll continue its other chores (eg. rebalancing). The use case here is I'd like to have onchain funds in the wallet (eg. for swaps if it's an exchange) but not spend it all to open new channels. The default value could be 0 meaning the current behaviour.

ZmnSCPxj:
Closes: #31 

@hosiawak can you please check if this is okay with you?  Also, please try to compile on your MacOS and FreeBSD as well if possible. [clboss-min-onchain.tar.gz](https://github.com/ZmnSCPxj/clboss/files/5577448/clboss-min-onchain.tar.gz)

### DIFF
--- a/Boss/Mod/ChannelCreationDecider.cpp
+++ b/Boss/Mod/ChannelCreationDecider.cpp
@@ -1,19 +1,25 @@
 #include"Boss/Mod/ChannelCreationDecider.hpp"
 #include"Boss/Msg/ChannelFunds.hpp"
+#include"Boss/Msg/ManifestOption.hpp"
+#include"Boss/Msg/Manifestation.hpp"
 #include"Boss/Msg/NeedsOnchainFunds.hpp"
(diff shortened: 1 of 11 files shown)


hosiawak:
@ZmnSCPxj Compiles on both systems and works as expected

ZmnSCPxj:
Thank you, merging.  Will make a 0.7B release soon.

PR #50, opened 2020-12-21
DnsSeed/get.cpp: Use `dig -v` to check for `dig` install.

### LINKED ISSUE #49: Delayed clightning startup when offline
When run on a system without WAN access, `clboss` adds about 18 seconds to service startup time (defined as the duration from starting the clightning process to the start of the RPC server). This delay is due to failed DNS resolution.

[clightning service output (gist)](https://gist.githubusercontent.com/erikarvstedt/2db1a591f1bc2b1ef0d92d59ea22f2dd/raw/7f6a6c1024ac6ecfc06d9ecc0c60ea67ebeda67d/clboss.txt)

#### clightning config file
```
network=bitcoin
bitcoin-datadir=/var/lib/bitcoind
proxy=127.0.0.1:9050
always-use-proxy=true
bind-addr=127.0.0.1:9735
bitcoin-rpcconnect=127.0.0.1
bitcoin-rpcport=8332
bitcoin-rpcuser=public
rpc-file-mode=0660
plugin=/nix/store/zd4y7s4vi43sjvmvyp6sv4zhc31rkk9l-clboss-0.10/bin/clboss
clboss-min-onchain=30000

disable-dns
bitcoin-rpcpassword=1s36RENe8gF7cYD8Zlm9
```

When `dig` is not in `PATH`, there is no delay: [clightning service output  (gist)](https://gist.githubusercontent.com/erikarvstedt/387f467bebb158126356c3dd89009795/raw/be830b68df38a5025affa132f8b80c189ea7a7b3/clboss-no-dig.txt).

This issue appeared when I added `clboss` to our [nix-bitcoin](https://github.com/fort-nix/nix-bitcoin) test suite, which runs bitcoin nodes including clightning in offline VMs.\
It would be great if `clboss` could handle missing WAN at startup without delaying the main clightning process.

ZmnSCPxj:
Fixes #49

@erikarvstedt can you check if this works?

Note that while it speeds up the pre-manifest times, due to `always-use-proxy` it will do a later check if `torify dig` works on each DNS seed (but as a background task after initialization).  This will again take 18 seconds or so with the WAN down.  If you are testing only "does it work" and you try to cleanly shut down `lightningd` (as opposed to `kill -s SIGINT`) immediately afterwards, the delay will still persist, this time on shutdown (because we have no facility that keeps track of forked processes and kills them when we want to die ourselves).

If so, maybe implementing a new `--clboss-disable-dnsseed` flag might do better, then you can add that in the config of your testcase.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,5 @@
+- Use `dig -v` to check for `dig` install, instead of `dig localhost`, as the latter may trigger a "real" lookup that will inevitably fail.
+
 0.10 Made of Explodium
 0.9A
 - Avoid `DELETE ... ORDER BY`, which might not be enabled on the SQLITE3 available on some systems.
--- a/DnsSeed/get.cpp
+++ b/DnsSeed/get.cpp
@@ -23,7 +23,7 @@ std::string enstring(a const& val) {
 namespace DnsSeed {
 
 Ev::Io<std::string> can_get() {
-	return Ev::runcmd("dig", {"localhost"}).then([](std::string _) {
+	return Ev::runcmd("dig", {"-v"}).then([](std::string _) {
 		return Ev::lift(std::string(""));
 	}).catching<std::runtime_error>([](std::runtime_error const& _) {
 		return Ev::lift(std::string(


erikarvstedt:
This works, thanks! In our case, there's no shutdown delay due to systemd's use of `SIGINT`.

PR #52, opened 2020-12-30
Implement IP Binning

### LINKED ISSUE #18: Avoid planning channels to nodes with similar IP addresses
If possible we should channel to nodes with different geographical locations based on published IP address, in order to avoid having all our channels going to various LNBIG nodes.

This requires a map of autonomous system numbers to IP addresses.

LNBIG is known to host multiple nodes in various locations but hopefully this reduces the probability of having all our channels to LNBIG.

For .onion:

* Give each unique .onion its own bucket?
* Put all .onion in a single unique bucket, but allow drawing more than one from that bucket?
  * Equivalently, create N buckets for .onions and throw onions randomly in one of the onion buckets.

Not sure at what point to implement this.  Probably best is to have the `ChannelCandidateInvestigator` add this as an investigation as well, and to penalize candidates that happen to have IPs similar to higher-scored candidates.

IRC discussion with @willcl-ark and @darosior :

```
<willcl_ark> zmnscpxj_: CLBOSS seemed to cope ok even during that crazy fee market. It did open two channels (when I had zero) at a rate of 250s/vB (two in one transaction), but then seemed to "wait" until fees were lower again to open a further 5. I suppose because of how agnositc it is, 5 out of 7 channels are to various LNBIG nodes which is probably not ideal, but difficult to automatically bucket peers by "likely same 
<willcl_ark> owner"
<zmnscpxj_> Yes, if a majority of your funds is outside of channels, CLBOSS will put as much as possible into channels even if feerates are high
<zmnscpxj_> but it tries to hold off as much as it can until feerates go low
<willcl_ark> That is pretty much exactly what I saw
<zmnscpxj_> works as designed then
<willcl_ark> :)
<zmnscpxj_> TheBlueMatt has suggested IP binning before (in the context of routing, but still...) so it might be useful to consider that as well
<zmnscpxj_> in order to avoid the "everryone is LNBIG in purgatory"
<willcl_ark> TBH, looks like the LNBIG nodes are hosted in different locations (on 1ML) so not sure that will help
<zmnscpxj_> awww crap
<willcl_ark> you could base it on Alias, but that's easy to break too
<zmnscpxj_> yes
<zmnscpxj_> and using IP binning does not help with .onion as well
<willcl_ark> actually I am not correct, all the LNBIG nodes I partnered with appear to be (Ashburn, VA, United States)
<willcl_ark> might be worth to have it "just in case", even if it's not perfect, I don't think it would make anything worse
<zmnscpxj_> okay, looks like IP binning might be useful at least to *reduce* the possibility
<willcl_ark> exactly
<zmnscpxj_> Dunno what to do about .onion though, oh well
<zmnscpxj_> And I have to wonder about IPv6 as well
<darosior> asmap ?
<zmnscpxj_> what is asmap?
<darosior> bitcoind's new technique to make sybil connections more costy but that may be nonsensical for your matter
<zmnscpxj_> references?
<darosior> https://github.com/bitcoin/bitcoin/issues/16599
<zmnscpxj_> yes, TheBlueMatt has also suggested this for LN routing as well
<zmnscpxj_> though he suggests all .onion should go into a single bin, which worries me
<willcl_ark> You could allow more from the tor bin, than other bins?
<zmnscpxj_> could be done as well.
```

ZmnSCPxj:
Fixes #18 

@hosiawak can I please ask a favor? This changeset adds code that uses OS interfaces, can you please compile this on MacOS and FreeBSD as well to check if the same OS interfaces exist there? ~~[clboss-ipbin.tar.gz](https://github.com/ZmnSCPxj/clboss/files/5752996/clboss-ipbin.tar.gz)~~ [clboss-ipbin.tar.gz](https://github.com/ZmnSCPxj/clboss/files/5753012/clboss-ipbin.tar.gz)

### DIFF
--- a/Boss/Mod/ChannelCreator/Manager.cpp
+++ b/Boss/Mod/ChannelCreator/Manager.cpp
@@ -13,8 +13,12 @@
 #include"Jsmn/Object.hpp"
 #include"Json/Out.hpp"
 #include"Ln/Amount.hpp"
+#include"Net/IPAddrOrOnion.hpp"
+#include"Net/IPBinnerBySubnet.hpp"
 #include"S/Bus.hpp"
+#include"Util/make_unique.hpp"
 #include<algorithm>
+#include<assert.h>
 #include<sstream>
 
 namespace {
@@ -36,6 +40,27 @@ bool plan_is_empty(std::map<Ln::NodeId, Ln::Amount> const& plan) {
 	});
 }
 
+Ev::Io<void> report_proposals( S::Bus& bus, char const* prefix
+			     , std::vector< std::pair<Ln::NodeId, Ln::NodeId>
+					  > const& proposals
+			     ) {
+	auto os = std::ostringstream();
+
+	auto first = true;
+	for (auto& p : proposals) {
+		if (first)
+			first = false;
+		else
+			os << ", ";
+		os << p.first;
+	}
+	return Boss::log( bus, Boss::Debug
+			, "ChannelCreator: %s: %s"
+			, prefix
+			, os.str().c_str()
+			);
+}
+
 }
 
 namespace Boss { namespace Mod { namespace ChannelCreator {
@@ -45,6 +70,12 @@ void Manager::start() {
 		     >([this](Msg::Init const& init) {
 		rpc = &init.rpc;
 		self = init.self_id;
+		reprioritizer = Util::make_unique<Reprioritizer>
+			( init.signer
+			, Util::make_unique<Net::IPBinnerBySubnet>()
+			, [this](Ln::NodeId n) { return get_node_addr(n); }
+			, [this]() { return get_peers(); }
+			);
 		return Ev::lift();
 	});
 	bus.subscribe<Msg::RequestChannelCreation
@@ -75,6 +106,8 @@ Manager::on_request_channel_creation(Ln::Amount amt) {
 			   + (double)info["num_active_channels"]
 			   ;
 		return investigator.get_channel_candidates();
+	}).then([this](std::vector<std::pair<Ln::NodeId, Ln::NodeId>> proposals) {
+		return reprioritize(std::move(proposals));
 	}).then([ this
 		, num_chans
 		, amt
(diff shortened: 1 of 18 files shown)


hosiawak:
@ZmnSCPxj  [FreeBSD output](https://gist.github.com/hosiawak/6cb34e87856c3ed41792ff5c3cba2174)

ZmnSCPxj:
Thank you @hosiawak, can you please try this instead? 
[clboss-ipbin.tar.gz](https://github.com/ZmnSCPxj/clboss/files/5753992/clboss-ipbin.tar.gz)

hosiawak:
@ZmnSCPxj fine on FreeBSD but on a Mac I [get this](https://gist.github.com/hosiawak/f42d643118e526ba3866884453f0edbd)

ZmnSCPxj:
Wut.

That code is not modified in this patchset.  Could you try this older repo on your Mac? https://github.com/ZmnSCPxj/clboss/releases/download/v0.10/clboss-0.10.tar.gz

hosiawak:
@ZmnSCPxj Yeah, same error on 0.10.

ZmnSCPxj:
Rebased after #54.

@hosiawak can you please try this on FreeBSD and MacOS? [clboss-ipbin.tar.gz](https://github.com/ZmnSCPxj/clboss/files/5754772/clboss-ipbin.tar.gz)

hosiawak:
@ZmnSCPxj Fine on both systems.

willcl-ark:
Just testing out this PR. I did notice something odd on Debian at startup, it complains in the log with 

``2020-12-31T11:42:09.560Z UNUSUAL plugin-clboss: DnsSeed: Cannot seed by DNS: `dig` command is not installed, it is usually available in a `dnsutils` package for your OS.``

...when my system _does_ have `dig` command installed.

Other than that, I had clboss disabled for a while (waiting for this patch) and startup (via systemd) kept failing. I think it was not returning a successful startup message to `systemctl` because that command would hang, and after a while clightning logs would stop and it would attempt restart.

I purged clboss data files from `~/.lightning/` and this seemed to sort out that issue.

Will report back later on channel creation and management more thoroughly; it appears to have only created two channels from ~0.057 ₿TC this time, vs the 6 or 7  it did last time, but neither are LNBIG!

ZmnSCPxj:
Aww man.  Maybe a side-effect of #50 ?  What do you get with `dig -v; echo $?` ?

willcl-ark:
> Aww man. Maybe a side-effect of #50 ? What do you get with `dig -v; echo $?` ?

```
will in ~ at will-nuc ➜ dig -v; echo $status
DiG 9.16.8-Debian
0
```

(because I'm using fish shell s/$?/$status)

ZmnSCPxj:
Fairly strange.  Can you file a new issue for this?  I cannot replicate it, lemme try to extract bits of CLBOSS that you can run in your system and which will print some more debug info for me to be able to see what is different in your system, this might take a few days however.  What is path of `dig` and did you execute `dig -v` as the same user C-Lightning is running as?

PR #54, opened 2020-12-30
Boss/Mod/InternetConnectionMonitor.cpp: Fix compile on MacOS.

ZmnSCPxj:
Hi @hosiawak, this fixes the compile problem you reported in https://github.com/ZmnSCPxj/clboss/pull/52#issuecomment-752442146

Could you please try to check compile on FreeBSD and MacOS? [clboss-for-vector-bool.tar.gz](https://github.com/ZmnSCPxj/clboss/files/5754355/clboss-for-vector-bool.tar.gz)


hosiawak:
@ZmnSCPxj  It's fine on both systems.

Issue #11, opened 2020-11-02
Modify fee by price theory


ZmnSCPxj:
Discussion with @whitslack here: https://old.reddit.com/r/Bitcoin/comments/jkibch/announcing_clboss_automated_clightning_node/gakg2e6/

In price theory:

* Given a particular demand profile for a product, there exists an optimal price where a supplier of the product maximizes its earnings.
* If the price is higher than the optimum price, fewer customers will buy, reducing the number of items sold.
* If the price is lower than the optimum price, more customers will buy, but the earning per customer is low, reducing earnings.

One thing we could do would be for a CLBOSS module to speculatively increase or decrease (at random) its fee modifier.  Then at each forward, it records how much it earns, and whether it is currently lower or higher than the base.  At some point, it changes what its "default" fee modifier is, depending on the recorded data.

ZmnSCPxj:
Implemented: https://github.com/ZmnSCPxj/clboss/commit/cf4e3fbafc07bfefaf451252077d1c47ef10061e

PR #55, opened 2020-12-31
Boss/Mod/EarningsTracker.cpp: Show offchain earnings v expenditure judgments in `clboss-status`.

### LINKED ISSUE #51: Possible to expose FundsMover via the RPC?
Is there a way to get FundsMover's actions somehow via the RPC (either identify self-payments somehow via the lightningd rpc or create clboss rpc that exposes this data) ?

plugin-clboss: FundsMover: Moved 201239852msat from xxx, getting 200678652msat to yyy, costing us 561200msat.

I'm trying to calculate operational costs of running a node so that we can subtract the total fees from the collected fees and get an idea if we're operating at a loss or profit and how it changes over time. So far I've got channel opening/closing fees done. FundsMover seems to be the next step and Boltz swap fees next after that.

ZmnSCPxj:
Fixes: #51

@hosiawak for your evaluation: 
[clboss-earnings-status.tar.gz](https://github.com/ZmnSCPxj/clboss/files/5757260/clboss-earnings-status.tar.gz)

This adds a new key `offchain_earnings_tracker` to the output of `clboss-status`.  This is an object, each key is a node that we have gotten earnings or spent `FundsMover` on.  This is just a dump of the `EarningsTracker` table in `data.clboss`.

Unlike what I said in #51 it turns out that I do not actually delete the rows in this table, ever, LOL, sorry.  *And* it shows that CLBOSS spends a lot more on rebalancing than it actually earns, oh no, woe is me, with anywhere from twice to 10x the expenditure vs earnings for each node.  Sigh.

### DIFF
--- a/Boss/Mod/EarningsTracker.cpp
+++ b/Boss/Mod/EarningsTracker.cpp
@@ -1,10 +1,12 @@
 #include"Boss/Mod/EarningsTracker.hpp"
 #include"Boss/Msg/DbResource.hpp"
 #include"Boss/Msg/ForwardFee.hpp"
+#include"Boss/Msg/ProvideStatus.hpp"
 #include"Boss/Msg/RequestEarningsInfo.hpp"
 #include"Boss/Msg/RequestMoveFunds.hpp"
 #include"Boss/Msg/ResponseEarningsInfo.hpp"
 #include"Boss/Msg/ResponseMoveFunds.hpp"
+#include"Boss/Msg/SolicitStatus.hpp"
 #include"Boss/concurrent.hpp"
 #include"Ev/Io.hpp"
 #include"S/Bus.hpp"
@@ -48,6 +50,13 @@ private:
 		bus.subscribe<Msg::RequestEarningsInfo
 			     >([this](Msg::RequestEarningsInfo const& req) {
 			return Boss::concurrent(request_earnings_info(req));
+		});
+
+		bus.subscribe<Msg::SolicitStatus
+			     >([this](Msg::SolicitStatus const& _) {
+			if (!db)
+				return Ev::lift();
+			return status();
 		});
 	}
 
@@ -212,6 +221,39 @@ private:
 				requester, node,
 				in_earnings, in_expenditures,
 				out_earnings, out_expenditures
+			});
+		});
+	}
+
+	Ev::Io<void> status() {
+		return db.transact().then([this](Sqlite3::Tx tx) {
+			auto fetch = tx.query(R"QRY(
+			SELECT node
+			     , in_earnings, in_expenditures
+			     , out_earnings, out_expenditures
+			  FROM "EarningsTracker"
+			     ;
+			)QRY").execute();
+
+			auto out = Json::Out();
+			auto obj = out.start_object();
+			for (auto& r : fetch) {
+				auto sub = obj.start_object(r.get<std::string>(0));
+				sub
+					.field("in_earnings", r.get<std::uint64_t>(1))
+					.field("in_expenditures", r.get<std::uint64_t>(2))
+					.field("out_earnings", r.get<std::uint64_t>(3))
+					.field("out_expenditures", r.get<std::uint64_t>(4))
+					;
+				sub.end_object();
+			}
+			obj.end_object();
+
+			tx.commit();
+
+			return bus.raise(Msg::ProvideStatus{
+				"offchain_earnings_tracker",
+				std::move(out)
 			});
 		});
 	}


hosiawak:
_And it shows that CLBOSS spends a lot more on rebalancing than it actually earns, oh no, woe is me, with anywhere from twice to 10x the expenditure vs earnings for each node_

Yeah, I noticed it too when calculating all the spent fees. I think the strategy to use competitive fees (especially _very_ competitive fee rates, the base fee calculation looks mostly fine to me) everywhere needs to be revisited. Perhaps it could use this rebalancing fee info to calculate "rebalancing fee rate" (RFR) for a channel and then use this RFR data when calculating out our target channel fee rate (add RFR to whatever fee rate it considers "competitive" now to make up for these rebalancing costs).

I think there are other optimisation techniques we could use here as well, need to think about it.

hosiawak:
Some data from my 2 month old node:

Fees collected: 42 373 sats
Fees spent (only channel opening/closing fees): 197 325 sats
Rebalancing fees: ???

So even without taking any rebalancing fees into account it's running at a loss.

hosiawak:
[My offchain earnings tracker log](https://gist.github.com/hosiawak/8a08ebe88e9a452ff18223924f5c2548)

Some channels are very active but it's more similar to a charity than a money making business for now :)

ZmnSCPxj:
> I think the strategy to use competitive fees (especially very competitive fee rates, the base fee calculation looks mostly fine to me) everywhere needs to be revisited.

In what direction?  Does it need to be higher or lower?  That is where #11 starts coming into consideration.  In theory, there is an optimum feerate for each of your channels.  *discovering* that feerate, now, how to go about it?

Anyway -- okay to merge for you?

hosiawak:
Ok to merge for me.

Feerate needs to be much higher imo. CLBOSS needs to be aware of its rebalancing fees. It starts with 0 rebalancing fees so it can use the current feerate calc initially but when the time goes by and rebalancing fees grow it should increase the feerate to make up for this maintenance cost.

If you look at my sample data you’ll find even 50x difference between the cost and earnings. Bringing this gap to 0 would be the first step so that by default it’s not „losing” sats. Once that’s done the next steps could involve better price theory optimization so that it starts „earning”. It could take channel opening/closing fees into account as well although they have less and less of an impact as time goes by.

In general the feerate clboss sets on my node is for balanced channels around 0.0001% and 0.01%. The channel balance is close to perfect (which is NOT common in today’s ln network) which means very reliable forwards both ways on one hand but extra maintenance cost on another and this reliability should be reflected in the price imo.

ZmnSCPxj:
Another thing to do would be to limit `InitialRebalancer` (this leads to lower balancing in general).  Also `FundsMover` itself deserves some optimization.  Lemme go dig that up as well.

btweenthebars:
So all the rebalancing cost is from the line "FundsMover: *** costing us" ? I've been running for 2 weeks with ignore on chain fund and the total cost is 4500sat while earning is about 1000sat.  There are some baffling rebalances that cost 400-800sat . Is it possible to know if which are regular rebalance or jitrebalance ? and for debugging purpose, shouldn't there be information about why it thinks it should rebalance on the log ?

ZmnSCPxj:
There are the following rebalancers, which you can see in the log:

* `InitialRebalancer` - Rebalances channels with a lot of capacity on our side, in favor of channels with low capacity on our side.  In recent versions this will be a truly "initial" rebalancer, i.e. it will not rebalance channels if we have already spent some amount on rebalancing.
* `EarningsRebalancer` - Rebalances funds out of channels where we have earned a lot of fees from incoming forwards, on the logic that in the future we can probably earn even more if we provide more incoming capacity on that channel.  In recent version, this prints the amusing message "refusing to throw good money after bad" all the time...
* `JitRebalancer` - Rebalances funds if we potentially cannot fulfill a forward due to lacking outgoing capacity on a channel.  In recent versions this also will not rebalance if we already spent a lot on funds on rebalancing the channels.

Prints are only seen on debug log, so if you want to see it in the logs, set `--log-level=debug:clboss`.

`FundsMover` is just a low-level helper that the above rebalancers order around.

It seems to me that currently the network has most forwarding nodes subsidizing the payments on Lightning, because after some time, none of the above rebalancers ever trigger due to hitting limits on how much they are willing to spend on rebalances.  This means the network is consistently undercutting the optimum feerate for forwarding nodes, probably due to current pathfinding algorithms always just concentrating on feerates and not on other parameters.

btweenthebars:
I see, it didn't rebalance for quite sometimes until I added new channels which triggers the InitialRebalance. I don't think the InitialRebalance makes sense on the current network. For example, I just participated in a LN+ 2M swap. The InitialRebalance kicked in to rebalance 1M, but instead of choosing the triangle path the I just made with LN+, it rebalanced to another node, and the cost was 1200 sat, instead of less than 100 in the right path. From what I have been observing with the current fee rate that clboss set, I might only get like 50sat back from the 1200sat investment.

ZmnSCPxj:
> instead of choosing the triangle path the I just made with LN+

Can you teach CLBOSS how to detect such a triangle path?

btweenthebars:
most liquidity swap, LN+/telegram has one incoming channel, and one outgoing channel with a similar size within the same time frame, (LN+ participants have to complete it within 2 days). Sometimes user might participate in a few swaps at once, but it's not difficult to figure out (try to getroute on every pair, etc). Sometimes, people would coordinate to set 0 fee to do rebalance, so the bot should sit tight until it makes sure nobody is keen to rebalance.

Later on the bot can use this routing information for future rebalancing. It will be interesting if clboss can participate in LN+ or similar platforms, automatically.

ZmnSCPxj:
Thanks, please file a new issue.  Are there API docs for this LN+?  A website?

I think what you want is, to have `InitialRebalancer` try to look for new channels with low outgoing capacity / high incoming capacity, and use those as the destination when it finds a channel with high outgoing capacity / low incoming capacity..

Possibly a redesign of `FundsMover` is in order, based on the observations:

* `InitialRebalancer` and `EarningsRebalancer` select a channel to move *from*, but could not care less where they move *to* (though `InitialRebalancer` would prefer that destinations do not retrigger `InitialRebalancer`, i.e. the funds moved *to* a small channel might be enough to exceed the 80% level that `InitialRebalancer` triggers on).
* `EarningsRebalancer` selects a channel to move *to*, but could not care less where the funds came *from*.

The current `FundsMover` requires that clients (the above rebalancers) specify *both* source *and* destination, but that could lead to rebalancers caring about the source and not the destination to specify a destination that is expensive to route to.  Arguably the `FundsMover` should be the one that selects a destination on behalf of rebalancers that care about the source but not the destination and vice versa.  Hmm.  Will take some time though, so please file a separate issue.

btweenthebars:
https://lightningnetwork.plus/ , you never heard ? It's responsible for 14k+ channels added since June. No API though, probably not at the stage where clboss can interact.

ZmnSCPxj:
Thanks, will check.  Please put two issues, one for the rebalancer/fundsmover improvement, the other for LN+ integration.  At the worst case we could always add simple HTML parsing, CLBOSS already depends on `libcurl` anyway so HTTPS is covered.

ZmnSCPxj:
LN+ looks like an implementation of this idea by some weird rando: https://zmnscpxj.github.io/offchain/cyclicsuperhubs.html

ZmnSCPxj:
Contacted LN+ admins, they are already planning an API of some sort.  Please make an issue so we can keep updated.

btweenthebars:
> LN+ looks like an implementation of this idea by some weird rando: https://zmnscpxj.github.io/offchain/cyclicsuperhubs.html

In case you missed it, there are also several telegram groups actively forming super hubs such as PLEBNET which turned into LN+ for better organize, Ring of Fire, Plebs2Salvador (they are making a ridiculous loop of 400+ node - for fun)

PR #57, opened 2021-01-01
Boss/Mod/FundsMover/Attempter.cpp: Be less willing to pay extra for randomized routes.

ZmnSCPxj:
Put as a PR so somebody else (@hosiawak?) can double-check my thinking.

`lightningd` actually has route randomization implemented in `getroute`, however its current `pay` implementation does not actually use it.  Route randomization gives privacy benefits, as (1) surveillors can trick naive `pay` implementations to going through their surveillance node by just advertising for tiny feerates (trading off fees for more information) and (2) always using the shortest route makes it trivial for surveillors to bound the possible destination of a payment to a small set of nodes.  Increased randomization also tends to mean we try out a wider variety of possible routes instead of retrying the same few ones.

Currently `FundsMover` uses rather high fuzzing of 99%.  This roughly translate to be willing to pay up to about double the fee compared to the cheapest route..

On the other hand, one can argue that a rebalance aka a self-payment "conveys no information", as you are paying yourself and therefore it does not link you to your actual economic partners.  It could theoretically leak an aggregate of other payments that went through your node, as rebalances are basically triggered due to forwards you facilitated unbalancing your channels, but aggregate information like that leaks very little information anyway.  So maybe lowering the route randomization parameter might be better; route randomization is needed for payments to other nodes, but maybe not so much for self-payments.

### DIFF
--- a/Boss/Mod/FundsMover/Attempter.cpp
+++ b/Boss/Mod/FundsMover/Attempter.cpp
@@ -18,6 +18,13 @@
 #include<string>
 #include<vector>
 
+namespace {
+
+auto initial_fuzzpercent = double(11.0);
+auto step_fuzzpercent = double(11.0);
+
+}
+
 namespace Boss { namespace Mod { namespace FundsMover {
 
 class Attempter::Impl : public std::enable_shared_from_this<Impl> {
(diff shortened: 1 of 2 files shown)


hosiawak:
@ZmnSCPxj As I understand the code FundsMover will use 11% route randomization and if that fails 0% ? If so it looks ok to me and ok to merge.

PR #59, opened 2021-01-18
Boss/Mod/FeeModderByExpenditures.cpp: Add module to try to increase earnings for high-expenditure channels.

### LINKED ISSUE #56: Fee Modder by Expenditures
We keep track of `FundsMover` expenditures anyway, but only use it to limit `JitRebalancer` to prevent targeted attacks if the attacker knows we use JIT rebalancing.  An idea from @hosiawak : https://github.com/ZmnSCPxj/clboss/pull/55#issuecomment-753238276

Here is a possible implementation:

* Compute `out_expenditures` divided by `out_earnings` for each channel as a `double`.  If this data is nonexistent, or `out_earnings` is zero, saturate to some maximum (maybe 100.0?).
*  Give that ratio as the multiplier returned by this `FeeModder`

For reference, the `FeeModder` system allows modules to provide a `double` that is multiplied to a computed basis fee.  This computed basis fee is the median of fees that other channels to that node have.

While this may be workable, we should note that if CLBOSS uptake is high, then since the basis we use is the fee that other channels charge, if the other nodes are also using CLBOSS this can lead to absurd situations where CLBOSS-managed nodes start bumping up channel feerates, then their competitors see the median as going higher and also bump up their channel feerates, etc.  Or similar absurd situations in the reverse direction.  So I'm wondering if the median-fee should instead be used as an initial fee when we have no information yet and to use #11 of some kind to change this initial feerate over time as we try to discover the optimum price based on our position in the network.

ZmnSCPxj:
Closes: #56

I have mixed feelings about this.  On my system, all channels will be hit by this, and while it *looks* like it increases the fee earnings, it might be better to actually tweak `FeeModderBySize` instead; in principle, we don't have to signal how much we have lost in rebalances.  In short, I think doing an across-the-board increase in fees we charge would be better than basing on our own local expenditures --- in all likelihood we need to bump up `FeeModderBySize` instead.

Lemme make an alternate patch and provide some packages, that will take several hours.

### DIFF
--- a/Boss/Mod/FeeModderByExpenditures.cpp
+++ b/Boss/Mod/FeeModderByExpenditures.cpp
@@ -1,0 +1,131 @@
+#include"Boss/Mod/FeeModderByExpenditures.hpp"
+#include"Boss/ModG/ReqResp.hpp"
+#include"Boss/Msg/ProvideChannelFeeModifier.hpp"
+#include"Boss/Msg/RequestEarningsInfo.hpp"
+#include"Boss/Msg/ResponseEarningsInfo.hpp"
+#include"Boss/Msg/SolicitChannelFeeModifier.hpp"
+#include"Boss/log.hpp"
+#include"Ev/Io.hpp"
+#include"S/Bus.hpp"
+#include"Util/make_unique.hpp"
+#include"Util/stringify.hpp"
+#include<math.h>
+#include<string>
+
+namespace {
+
+/* Base to use for logarithm.  */
+auto constexpr log_base = double(10);
+/* Maximum multiplier.  */
+auto constexpr max_multiplier = double(5.0);
+
+}
+
+namespace Boss { namespace Mod {
+
+class FeeModderByExpenditures::Impl {
+private:
+	S::Bus& bus;
+	Boss::ModG::ReqResp< Msg::RequestEarningsInfo
+			   , Msg::ResponseEarningsInfo
+			   > earnings_rr;
+
+	void start() {
+		bus.subscribe< Msg::SolicitChannelFeeModifier
+			     >([this](Msg::SolicitChannelFeeModifier const& _) {
+			return bus.raise(Msg::ProvideChannelFeeModifier{
+				[this](Ln::NodeId n, std::uint32_t _, std::uint32_t __) {
+					return modify_fee(n);
+				}
+			});
+		});
+	}
+
+	Ev::Io<double>
+	modify_fee(Ln::NodeId const& n) {
+		return Ev::lift().then([this, n]() {
+			return earnings_rr.execute(Msg::RequestEarningsInfo{
+				nullptr, n
+			});
+		}).then([this, n](Msg::ResponseEarningsInfo e) {
+			auto msg = std::string();
+			auto factor = double();
+			decide_fee_factor(msg, factor, e.out_earnings, e.out_expenditures);
+
+			return Boss::log( bus, Debug
+					, "FeeModderByExpenditures: %s "
+					  "(earn = %s, spend = %s): %s"
+					, Util::stringify(n).c_str()
+					, Util::stringify(e.out_earnings).c_str()
+					, Util::stringify(e.out_expenditures).c_str()
+					, msg.c_str()
+					).then([factor]() {
+				return Ev::lift(factor);
+			});
+		});
+	}
+
+	void decide_fee_factor( std::string& msg
+			      , double& factor
+			      , Ln::Amount earnings
+			      , Ln::Amount expenditures
+			      ) {
+		if (earnings >= expenditures) {
+			msg = "Channel not at a loss, factor = 1.";
+			factor = 1.0;
+			return;
+		}
+
+		/* Avoid divide by 0.  */
+		if (earnings == Ln::Amount::msat(0)) {
+			msg = std::string("No earnings, factor = max = ")
+			    + Util::stringify(int(max_multiplier))
+			    + "."
+			    ;
+			factor = max_multiplier;
+			return;
+		}
+
+		auto ratio = expenditures / earnings;
+		factor = 1.0 + round(::log(ratio) / ::log(log_base));
+		msg = std::string("spend / earn = ")
+		    + Util::stringify(ratio)
+		    + ", factor = "
+		    ;
+		if (factor <= max_multiplier)
+			msg += Util::stringify(int(factor))
+			     + "."
+			     ;
+		else {
+			factor = max_multiplier;
+			msg += std::string("max = ")
+			     + Util::stringify(int(factor))
+			     + "."
+			     ;
+		}
+	}
+
+public:
+	Impl() =delete;
+	Impl(Impl&&) =delete;
+	explicit
+	Impl( S::Bus& bus_
+	    ) : bus(bus_)
+	      , earnings_rr( bus_
+			   , [](Msg::RequestEarningsInfo& m, void* p) {
+				m.requester = p;
+			     }
+			   , [](Msg::ResponseEarningsInfo& m) {
+				return m.requester;
+			     }
+			   )
+	      { start(); }
+};
+
+FeeModderByExpenditures::FeeModderByExpenditures(FeeModderByExpenditures&&) =default;
+FeeModderByExpenditures::~FeeModderByExpenditures() =default;
+
+FeeModderByExpenditures::FeeModderByExpenditures(S::Bus& bus)
+	: pimpl(Util::make_unique<Impl>(bus)) { }
+
+}}
--- a/Boss/Mod/FeeModderByExpenditures.hpp
+++ b/Boss/Mod/FeeModderByExpenditures.hpp
@@ -1,0 +1,33 @@
+#ifndef BOSS_MOD_FEEMODDERBYEXPENDITURES_HPP
+#define BOSS_MOD_FEEMODDERBYEXPENDITURES_HPP
+
+#include<memory>
+
+namespace S { class Bus; }
+
+namespace Boss { namespace Mod {
+
+/** class Boss::Mod::FeeModderByExpenditures
+ *
+ * @brief Modify fees according to expenditures.
+ * If we have lots of expenditures, increase the fee
+ * we ask for.
+ */
+class FeeModderByExpenditures {
+private:
+	class Impl;
+	std::unique_ptr<Impl> pimpl;
+
+public:
+	FeeModderByExpenditures() =delete;
+
+	FeeModderByExpenditures(FeeModderByExpenditures&&);
+	~FeeModderByExpenditures();
+
+	explicit
+	FeeModderByExpenditures(S::Bus&);
+};
+
+}}
+
+#endif /* !defined(BOSS_MOD_FEEMODDERBYEXPENDITURES_HPP) */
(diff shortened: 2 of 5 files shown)


ZmnSCPxj:
*  [clboss-by-expenditures.tar.gz](https://github.com/ZmnSCPxj/clboss/files/5828368/clboss-by-expenditures.tar.gz)

@hosiawak could you please evaluate this in comparison to #61 ?

If you need a control that contains neither patch, the 0.11A release should be good: https://github.com/ZmnSCPxj/clboss/releases/tag/0.11A

hosiawak:
@ZmnSCPxj Ok, I'll take a snapshot of the current state of clboss:

- its earnings/expenditures report
- lightningd channel fees
- anything else ?

Then run this PR for a week and do the above then go back to the control for a day and then FeeModderBySize.
And finally run the numbers and try to come up with some conclusion :)

If you have other ideas please let me know.

ZmnSCPxj:
Primary goal is to increase earnings always, so the `getinfo` `fees_collected_msat` is probably our best bet for judging that.  We just really need the deltas.

You could also try grepping the `ForwardFeeMonitor` and see how varied it is, maybe a `grep "ForwardFeeMonitor" logfile | wc` would do.  That assumes you have debug-level logs though --- it only prints at debug.  Or you could modify the CLBOSS source code to change it to Info level.  I notice that fewer payments passed through my node, and along fewer in/out paths, when I enabled by-expenditures.

Problem is that this takes weeks of data-gathering to do... sigh.

hosiawak:
@ZmnSCPxj I've had this version running for the past 6 hours or so. I've seen FeeModderBySize in the log but FeeModderByExpenditure has been quiet. When should I expect it to say something (roughly) ?

ZmnSCPxj:
Actually you should.... hmmm.

```
2021-01-18T00:01:22.968Z DEBUG   plugin-clboss: FeeModderByExpenditures: ${PEER} (earn = 1680628msat, spend = 7853457msat): spend / earn = 4.67293, factor = 2.
```

All the `FeeModder`s should be talking at around the same time, since they have the same trigger.

Lemme check what happened....

ZmnSCPxj:
Well, I:

* Downloaded the file: https://github.com/ZmnSCPxj/clboss/files/5828368/clboss-by-expenditures.tar.gz
* Untarred, `./configure && make -j3`
* Modified `lightning/config` to point to the `clboss-by-expenditures/clboss`.
* Restarted C-Lightning.

And got:

```
2021-01-19T23:48:59.540Z DEBUG   plugin-clboss: FeeModderByExpenditures: ${PEER} (earn = 1680628msat, spend = 7853457msat): spend / earn = 4.67293, factor = 2.
```

hosiawak:
@ZmnSCPxj FeeModderBySize uses INFO level but ByExpenditures and ByBalance use DEBUG:

Screenshot 2021-01-20 at 01 09 19

ZmnSCPxj:
Ah, I see, how inconsistent, LOL.  I will make all the fee modders `Info` once we have gotten a 59/61/neither decision.

hosiawak:
After switching to this modder I'm noticing similar number of forwards (previous period: 330 this period: 331) and less collected fees (previous period: 7k sats this period: 4k sats) which doesn't make sense on the first glance so still investigating.

Among all 60 channels I didn't notice any dramatic fee increase except for 1 or 2 channels. I'll collect full week of data on Sun and run the numbers then.

ZmnSCPxj:
Thank you.  I saw a mild increase in total earnings with this patch compared to prior, but with fewer channels participating in forwards.

Confounding variables include actual differences in Lightning usage by others and changing network graph.

ZmnSCPxj:
In addition, the linked tarball includes 0.11A, which includes changes to reduce rebalancer aggression because the overaggressive rebalancers just lose money.  So it might be fairer to compare to 0.11A rather than your previous experience prior to this patch i.e. we should compare earnings minus expenditures.  The 0.11A changes are definitely desirable anyway, since it limits your losses.

hosiawak:
Base fee and fee rates: control vs the expenditure mod running for a week.

`
Channel | Control |   | Expenditure mod |   | Base fee change | Fee rate change
-- | -- | -- | -- | -- | -- | --
  | Fee base | Fee rate | Fee  base | Fee rate |   |  
C1 | 11581 | 1158 | 21137 | 10568 | 9556 | 9410
C2 | 1391 | 973 | 1402 | 981 | 11 | 8
C3 | 3041 | 6 | 9064 | 91 | 6023 | 85
C4 | 506 | 1 | 1973 | 99 | 1467 | 98
C5 | 469 | 5 | 23 | 11568 | -446 | 11563
C6 | 1362 | 14 | 3411 | 34 | 2049 | 20
C7 | 1231 | 1 | 1309 | 1 | 78 | 0
C8 | 1 | 1 | 1 | 318 | 0 | 317
C9 | 24 | 1 | 494 | 5 | 470 | 4
C10 | 1332 | 1 | 4989 | 50 | 3657 | 49
C11 | 3192 | 638 | 3192 | 638 | 0 | 0
C12 | 14 | 14 | 4 | 4 | -10 | -10
C13 | 2 | 2 | 11 | 11 | 9 | 9
C14 | 1712 | 2 | 8880 | 53 | 7168 | 51
C15 | 318 | 1 | 1833 | 18 | 1515 | 17
C16 | 1762 | 2 | 11802 | 12 | 10040 | 10
C17 | 6554 | 8 | 8784 | 103 | 2230 | 95
C18 | 1 | 1 | 8385 | 419 | 8384 | 418
C19 | 359 | 1 | 1437 | 1 | 1078 | 0
C20 | 1436 | 1 | 4309 | 267 | 2873 | 266
C21 | 352 | 1 | 1410 | 352 | 1058 | 351
C22 | 1 | 1 | 4 | 46 | 3 | 45
C23 | 1 | 1 | 15 | 15 | 14 | 14
C24 | 1 | 1 | 2 | 420 | 1 | 419
C25 | 438 | 4 | 964 | 10 | 526 | 6
C26 | 1 | 43 | 39 | 258 | 38 | 215
C27 | 3163 | 95 | 2268 | 567 | -895 | 472
C28 | 626 | 1 | 12551 | 3138 | 11925 | 3137
C29 | 990 | 35 | 1500 | 75 | 510 | 40
C30 | 1 | 1 | 2 | 533 | 1 | 532
C31 | 1 | 12 | 4 | 7 | 3 | -5
C32 | 266 | 1 | 799 | 659 | 533 | 658
C33 | 480 | 1 | 16612 | 17 | 16132 | 16
C34 | 10434 | 104 | 13512 | 14 | 3078 | -90
C35 | 226 | 11 | 620 | 30 | 394 | 19
C36 | 1628 | 2 | 4504 | 45 | 2876 | 43
C37 | 1635 | 2 | 8202 | 82 | 6567 | 80
C38 | 3150 | 3 | 12739 | 13 | 9589 | 10
C39 | 1032 | 595 | 2065 | 1189 | 1033 | 594
C40 | 247 | 2 | 793 | 396 | 546 | 394
C41 | 1 | 1 | 3 | 870 | 2 | 869
C42 | 2 | 2 | 9 | 9 | 7 | 7
C43 | 26 | 26 | 77 | 77 | 51 | 51
C44 | 1021 | 10 | 59656 | 60 | 58635 | 50
C45 | 2584 | 258 | 2457 | 546 | -127 | 288
C46 | 2058 | 2 | 18395 | 18 | 16337 | 16
C47 | 1439 | 1 | 3769 | 4 | 2330 | 3
C48 | 978 | 1 | 3879 | 4 | 2901 | 3
C49 | 1660 | 2 | 4979 | 1245 | 3319 | 1243
C50 | 553 | 39 | 426 | 128 | -127 | 89
C51 | 1 | 1 | 7 | 1665 | 6 | 1664
C52 | 5351 | 5 | 5351 | 5 | 0 | 0
C53 | 161 | 1 | 599 | 1 | 438 | 0
C54 | 2472 | 47 | 7626 | 1906 | 5154 | 1859
C55 | 2750 | 275 | 10144 | 2536 | 7394 | 2261
C56 | 82 | 43 | 82 | 43 | 0 | 0
C57 | 1000 | 10 | 1000 | 10 | 0 | 0
C58 | 6656 | 33 | 24189 | 242 | 17533 | 209
C59 | 466 | 233 | 466 | 233 | 0 | 0
C60 | 554 | 6 | 4911 | 491 | 4357 | 485
C61 | 1 | 1 | 1 | 21 | 0 | 20
C62 | 1 | 1 | 1 | 238 | 0 | 237
C63 | 841 | 411 | 2151 | 1052 | 1310 | 641
C64 | 1497 | 15 | 745 | 26 | -752 | 11
  |   |   |   | AVG | 3576 | 615
  |   |   |   | MEDIAN | 530 | 478

`

Some conclusions based only on fee changes:

1. Even though in some cases there's been a dramatic X fold increase in fee rates (eg. C4 - 99x, C20 - 267x) because we're staring from a very low fee 1msat per millionth the end result is still a low fee
2. The same applies to base fees although to a lesser extent (eg. C9 from 24 to 494)
3. Average base fee change: 3576 msat
4. Median base fee change: 530 msat
5. Average fee rate change: 615 msat
6. Median fee rate change: 478 msat

hosiawak:
Control number of settled forwards: 547
Control fees earned: 9,719,151 msat
Control value forwarded: 151,585,252,240 msat

Exp mod number of settled forwards: 381
Exp mod fees earned: 4,382,770 msat
Exp mod value forwarded: 89,348,547,253 msat

hosiawak:
Looks like running with this mod is worse than running without it in terms of earned fees and value forwarded. I'll go and test https://github.com/ZmnSCPxj/clboss/pull/61 now. Give it another week and see what happens.

Do you think I need to run 0.11A first for a day or so to give it some time to figure out the initial fees again or does it not matter ?

As a side note I noticed a need for another knob: disabling fee setting altogether by clboss. I still want to have clboss running but would like to experiment with setting fees using other strategies/plugins. Do you think it would be possible to add @ZmnSCPxj ?

hosiawak:
https://github.com/ZmnSCPxj/clboss/pull/61#issuecomment-776842402

ZmnSCPxj:
> As a side note I noticed a need for another knob: disabling fee setting altogether by clboss. I still want to have clboss running but would like to experiment with setting fees using other strategies/plugins. Do you think it would be possible to add @ZmnSCPxj ?

Master now has a `clboss-unmanage` command that lets you tell CLBOSS not to manage lnfees for specific nodes, e.g.

    lightning-cli clboss-unmanage ${NODEID} lnfee

See the `README.md` file.

It is a bit unwieldy since it works on a per-node level, i.e. you have to disable `lnfee` management for each node.  I will work on allowing `'*'` for the nodeid to completely disable management for all nodes.

hosiawak:
@ZmnSCPxj clboss-unmanage lnfee is great.

Would it be difficult to add clboss-unmanage balance as well ?

The use case is I provide liquidity to some merchants using @lightningliquiditybot and don't want clboss to immediately start sending sats on the newly opened channels (it will try to balance a newly opened channel).

What would be even more awesome is ability to set the ideal channel balance (eg. us | balanced | them) and clboss would then strive to achieve the desired balance. By default clboss would use `balanced` for all channels but users could change it to `us` (when providing liquidity to merchants) or `them` (when being a merchant).

ZmnSCPxj:
Superseded by `FeeModderByPriceTheory`.

PR #61, opened 2021-01-18
Boss/Mod/FeeModderBySize.cpp: Tweak parameters.

### LINKED ISSUE #56: Fee Modder by Expenditures

ZmnSCPxj:
Alternative to #59

Closes: #56 

This basically just increases the fee that `FeeModderBySize` will make.  In general it doubles the fee, *almost* doubles it if the node is "small" by its size measure.

Will go build a package, please wait....

### DIFF
--- a/Boss/Mod/FeeModderBySize.cpp
+++ b/Boss/Mod/FeeModderBySize.cpp
@@ -45,6 +45,10 @@ size, and comparing ourselves with the competitors.
 
 namespace {
 
+auto constexpr min_multiplier = double(0.9);
+auto constexpr max_multiplier = double(32.0);
+auto constexpr lowest_max_multiplier = double(4.0);
+
 /* Given the number of nodes that are lower-capacity than us,
  * the number who are higher-capacity, and the total number of
  * competitors, what is the multiplier on the feerates we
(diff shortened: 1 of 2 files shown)


ZmnSCPxj:
* [clboss-by-size-tweak.tar.gz](https://github.com/ZmnSCPxj/clboss/files/5828362/clboss-by-size-t#59 .tar.gz)

@hosiawak could you evaluate this in comparison to #59?

hosiawak:
@ZmnSCPxj I'll rebase this on top of the current master and evaluate again. Basically I've been evaluating these modders without a running rebalancer since 19th Jan so the results and conclusions are completely wrong. I'll come back in a couple of weeks with some results.

ZmnSCPxj:
Closed due to staleness.

PR #62, opened 2021-01-19
Tweak channel fees to make base higher and proportional lower

ZmnSCPxj:
This follows the "forwarder is payment aggregator" argument, which is that the forwarding node is in the business of aggregating multiple small payments via the shortest path (through their own node) and using longer paths to ensure their shortest-path channels remain balanced.  In order to earn, the forwarding node depends on the base fee and not the proportional fee. https://github.com/ZmnSCPxj/clboss/issues/56#issuecomment-761904848

* Tarball: [clboss-by-forwarder.tar.gz](https://github.com/ZmnSCPxj/clboss/files/5834416/clboss-by-forwarder.tar.gz)

This is not an alternate to #59  or #61 , I will likely merge this in regardless, but it might be useful for others to actually have a look at it and check if the logic in the "forwarder is payment aggregator" argument makes sense.

### DIFF
--- a/Boss/Mod/ChannelFeeManager.cpp
+++ b/Boss/Mod/ChannelFeeManager.cpp
@@ -1,6 +1,7 @@
 #include"Boss/Mod/ChannelFeeManager.hpp"
 #include"Boss/Msg/PeerMedianChannelFee.hpp"
 #include"Boss/Msg/ProvideChannelFeeModifier.hpp"
+#include"Boss/Msg/ProvideChannelFeeModifierDetailed.hpp"
 #include"Boss/Msg/SetChannelFee.hpp"
 #include"Boss/Msg/SolicitChannelFeeModifier.hpp"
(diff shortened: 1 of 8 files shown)


hosiawak:
@ZmnSCPxj I won't complain but sooner or later someone will that micropayments are no longer possible. I think eventually clboss will need a few knobs for setting various fee strategies to satisfy this and other needs.

I'll give this a test too and report back once I have something.

ZmnSCPxj:
Due to the modular nature of CLBOSS, it may be possible to just include an option to outright prevent particular modules from being instantiated.  Of course, the relative lack of C++ introspection makes that difficult.

In any case I intend to merge this in later after we have a 59/61/neither decision.

ZmnSCPxj:
Closing since this may not make much sense.

I may change channel fee system after 0.12.

PR #63, opened 2021-01-23
Have explicit .onion Boltz

### LINKED ISSUE #45: Support .onion fallback for Boltz-compatible swaps
Recently, `boltz.exchange` had an expired SSL certificate which prevented CLBOSS from being able to reverse the polarity of our channels.

A possible workaround would have been to use the .onion, `boltzzzbnus4m7mta3cxmflnps4fp7dueu2tgurstbvrbt6xswzcocyd.onion/`, since this self-certifies.

What we could do would be to support using a "main" name and a fallback name, with the fallback name being an .onion.

ZmnSCPxj:
Fixes: #45 

This adds support for fallback onion addresses in case the clearnet addresses are unavailable.

This also adds @AutonomousOrganization Boltz instance as well.

Untested!  I am running evaluation for 59/61/neither now, so unable to test this yet, bleah.

### DIFF
--- a/Boltz/Connection.hpp
+++ b/Boltz/Connection.hpp
@@ -1,59 +1,27 @@
 #ifndef BOLTZ_CONNECTION_HPP
 #define BOLTZ_CONNECTION_HPP
 
-#include<memory>
-#include<stdexcept>
-#include<string>
-
-namespace Ev { template<typename a> class Io; }
-namespace Ev { class ThreadPool; }
-namespace Jsmn { class Object; }
-namespace Json { class Out; }
+#include"Boltz/Detail/NormalConnection.hpp"
 
 namespace Boltz {
 
 /** class Boltz::Connection
  *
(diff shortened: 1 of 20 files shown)


ghost:
Hey this is cool and btw great work on budgeting the balancing.  Pulled and started running this branch  to see if I can get a swap, but it started to swap with the original boltz server, so I made a small modification to make my boltz server the only choice. Left it run overnight, did  not make contact.

3nprob:
Thanks so much fo ryour work on this @ZmnSCPxj 

While this seems to solve the specific issue at this point in time, wouldn't it make more sense to add a configuration variable for the Boltz API endpoint and node? As in https://github.com/BoltzExchange/channel-creation-plugin

That would also:
* Allow running using your own custom one (either for local dev testing or for alternative locations)
* Remove the need to keep this plugin up to date if/when endpoints change

PR #74, opened 2021-08-27
Configure min|max channel size

willcl-ark:
This PR aims to allow users to configure {min|max} channel sizes that CLBOSS should open. Motivation is wanting to permit CLBOSS to open larger channels.

This is presented as an advanced configuration option passed to `lighthingd` and if unset the current defaults remain.

Currently untested (CLBOSS doesn't seem to like my tiny regtest network of 2 nodes!) and no tests written. These may be added to this PR shortly if there is interest.

@ZmnSCPxj interested in your feedback on whether you want to see advanced config options like this in CLBOSS at all? The other alternative is to add some interior logic which opens larger channels if there is a large wallet balance in `lightningd`, or perhaps even uses a target "number of channels" to open e.g. ~5 channels. If you drop 0.1 BTC in, there won't be any wumbo channels, but if you drop 1 BTC in it might open one or two wumbo channels?

The final alternative is to do nothing at all, and CLBOSS will simply keep making many channels under the previous soft channel limit of 0.16...btc.

### DIFF
--- a/Boss/Mod/ChannelCreator/Manager.cpp
+++ b/Boss/Mod/ChannelCreator/Manager.cpp
@@ -5,6 +5,9 @@
 #include"Boss/Mod/ChannelCreator/RearrangerBySize.hpp"
 #include"Boss/Mod/Rpc.hpp"
 #include"Boss/Msg/Init.hpp"
+#include"Boss/Msg/ManifestOption.hpp"
+#include"Boss/Msg/Manifestation.hpp"
+#include"Boss/Msg/Option.hpp"
 #include"Boss/Msg/RequestChannelCreation.hpp"
 #include"Boss/Msg/SolicitChannelCandidates.hpp"
 #include"Boss/concurrent.hpp"
@@ -26,8 +29,8 @@
 namespace {
 
 /* Minimum and maximum channel size.  */
-auto const min_amount = Ln::Amount::btc(0.005);
-auto const max_amount = Ln::Amount::sat(16777215);
+auto const default_min_amount = Ln::Amount::btc(0.005);
+auto const default_max_amount = Ln::Amount::sat(16777215);
 /* If it is difficult to fit the amount among multiple
  * proposals, how much should we target leaving until
  * next time?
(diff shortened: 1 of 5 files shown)


willcl-ark:
This implementation is not yet complete, you'd ideally like to guarantee that your counterparty _also_ has some large-capacity channels so that your liquidity is not locked up wastefully.

As we already check for nodes with "too little capacity to be worth channeling with", this would want to be extended to consider increased capacities of potential wumbo peers.

Boss/Mod/ChannelCreator/Manager.cpp
```diff
@@ -26,8 +29,8 @@
 namespace {
 
 /* Minimum and maximum channel size.  */
-auto const min_amount = Ln::Amount::btc(0.005);
-auto const max_amount = Ln::Amount::sat(16777215);
+auto min_amount = Ln::Amount::btc(0.005);
```
ZmnSCPxj: Preferably we should instead rename these constants to `default_min_amount` and `default_max_amount`, then add `Ln::Amount min_amount` and  `Ln::Amount max_amount` to `class Manager` (and make `custom_min_amount` local to the triggered functions, or even just use the `class Manager` variables directly).

As a general principle we should avoid *mutable* (i.e. non-`const`/`constexpr`) [globals](https://wiki.c2.com/?GlobalVariablesConsideredHarmful) (even if only file-local scope, if its lifetime technically exceeds the life `main`, it is global), even if we know they will be used in classes that in practice will be single instance.

ZmnSCPxj (changes requested):

Boss/Mod/ChannelCreator/Manager.cpp
willcl-ark: OK I think I've reworked this to use the `class Manager` variables directly, as that seemed cleanest.

There's no sanity checking currently, and it's probably possible to silly things like configure `max < min` or to set these to other values which do not make sense.

Boss/Mod/ChannelCreator/Manager.cpp
```diff
@@ -68,7 +71,65 @@ Ev::Io<void> report_proposals( S::Bus& bus, char const* prefix
...
+                "Maximum channel size CLBOSS will create. "
+                "Set to an integer number of satoshis. "
+                "default=16777215"
+        });
+    });
+    bus.subscribe<Msg::Option
+    >([this, &custom_max_amount](Msg::Option const& o) {
+        if (o.name != "clboss-max-channel-size")
+            return Ev::lift();
+        custom_max_amount = Ln::Amount::sat((double) o.value);
```
ZmnSCPxj: Why not just:

```c++
max_amount = Ln::Amount::sat((double) o.value);
```

and remove the `custom_`\* variiables completely?

Also I prefer the casting style `double(o.value)` over `(double) o.value`.

willcl-ark: That makes a lot more sense. Now updated (and fixed indentation inside the lambdas to match your style, I think!)

Boss/Mod/ChannelCreator/Manager.cpp
```diff
@@ -68,7 +71,55 @@ Ev::Io<void> report_proposals( S::Bus& bus, char const* prefix
...
+
+    bus.subscribe<Msg::Manifestation
+            >([this](Msg::Manifestation const& _) {
+        return bus.raise(Msg::ManifestOption{
+                "clboss-min-channel-size"
+                , Msg::OptionType_Int
+                , Json::Out::direct(min_amount.to_sat())
+                , "Minimum channel size CLBOSS will create. "
+                  "Set to an integer number of satoshis. "
+                  "default=500000"
```
ZmnSCPxj, line 85: `ManifestOption` accepts a `std::string` in this position, so it should be possible to get this from the `default_max_amount` value with some amount of `+` and `std::string` concatenation.

ZmnSCPxj (commented):
Code itself looks fairly good; what are your further plans?

Re: indent: I use `tabstop=8` and actual tab chars, not spaces.

jonasnick:
Just want to state that I'd like to open larger channels. So this would be useful.

willcl-ark:
OK I got #reckless and ran this patch on mainnet where it successfully opened channels only between the sizes specified in the config file 👍🏼

@ZmnSCPxj I wonder if we want to as part of this PR bump the default max size of a channel to somewhere in the range 1,000,000 - 2,000,000 satoshis, as this is what certain services (e.g. LNRouter, maybe also lightning.terminal -- not sure) are "expecting" for a "good" routing channel?

ZmnSCPxj:
> default max size of a channel to somewhere in the range 1,000,000 - 2,000,000 satoshis

You want to lower the MAX size from 16,777,215sats to just 2,000,000 sats?  That seems restrictive --- I would expect such secondary tools to actually prefer larger channels?

willcl-ark:
Youre right I have it backwards-- I would not like to do that! But in perhaps default `min_amount` should be increased to 2,000,000 or 1,000,000?

Opening channels of a lesser size than that does not seem optimal nowadays.

ZmnSCPxj:
1,000,000sats for now please.  An issue is that triggering of channel creation may also need rethinking as well due to this minimum channel size --- `ChannelCreationDecider` currently triggers on 1,000,000 sats onchain, theminimum channel size may need to affect that module too.

It might be better to move those settings to `ChannelCreationDecider`, and have it trigger on `min_channel + reserve` instead.  And send the `min_channel` and `max_channel` settings via the `Boss::Msg::RequestChannelCreation` message (i.e. modify the message to include those fields) to `ChannelCreator`, and have its manager extract the settings from the message.

ZmnSCPxj:
We probably also want to impose 500000sat <= `min_channel` < `max_channel` <= 16777215sat.

ZmnSCPxj:
Closing since similar behavior already exists

Issue #75, opened 2021-10-12
Sparse records of swaps and FundsMover pays


tim-tx:
I started a new node with clboss a few days ago. I am trying to "audit" clboss and do double-entry accounting for all movement of funds on the node.

It was a bit difficult to find which pay resulted in new onchain funds in the event of a swap based just on the system logs. I noticed that a `SwapManager` record (that links `address` and `payment_hash`) in clboss's sqlite db gets wiped after a successful swap. I wonder if in the future this information might be preserved somewhere? Otherwise the process is to look at `listpays` and `listfunds true` and search for entries of approximately equal amount (differing by unknown fees). Looking in the system logs for `SwapManager` entries was confusing also due to disparate begin and end amounts for the same uuid:
```
plugin-clboss: SwapManager: Swap 7eebf286b148f88d7ecef301818b4ef7 started for 1923741188msat.
plugin-clboss: SwapManager: Swap 7eebf286b148f88d7ecef301818b4ef7 completed with 752127000msat onchain.
```
I haven't peeked at this portion of the code but I'm guessing "started for" is the target onchain amount, and "completed with" is the amount obtained in the swap, presumably meeting the target with whatever we already had onchain. 

`FundsMover` seems to wipe pay records as well, not just from the clboss db but from c-lightning. I can understand wanting not to clutter pay records with automated activity, but I wonder if in the future this information might be saved also. Otherwise I have to grep system logs for `FundsMover` and parse entries like this:
```
plugin-clboss: FundsMover: Moved 187404944msat from 033b90ec27e6d8140f9d56910312815fc8d166d2bb93a2fae6d35c9e84032dbb62, getting 187081604msat to 0337a42a4384c331258f7fa53697f27dbc1596a689fe77923d31b8e7a6f5960929, costing us 323340msat.
```

Finally, in my personal accounting I find I am in agreement with `utxo_amount` from `lightning-cli summary`, but my own accounting exceeds `avail_out` by ~60k sats. (It's a whole number of sats, I don't know if this is a coincidence.) Activity accounted for into and out of channels was

  1. Opening channels
  2. Pays for swaps
  3. A single user-initiated test pay of 10 sats plus 2msat fee
  4. `FundsMover` fees parsed from logs
  5. Fee earnings from forwards

Every entry in `listpays` with `status == "complete"` and every entry in `listtransactions` and `listfunds` is accounted for. Is there some other source of fees or debit of funds I might be missing, other than the list above?

*TL;DR*
[...]

PR #77, opened 2021-11-15
Track offchain-to-onchain swaps

ZmnSCPxj:
Part of #75 

Untested!  Will evaluate this over the next few days.


ZmnSCPxj:
Looks good, but I added an `amount_lost` field per entry, will merge as soon as CI passes.

PR #82, opened 2021-11-27
ChannelFinderByPopularity becomes aggressive when owned funds jumps significantly

### LINKED ISSUE #81: The bigger the node, the bigger the channels, the smaller the peers.
Once reaching ~35Msat local capacity, clboss starts to open larger channels, but is also choosing smaller peers. 
As a result a huge percent of total liquidity is allocated to very small peers.
For example, the latest automatic channel creation was 6M+ for a peer of 28M total capacity. 

Expected behavior :
Consider the peer capacity at channel size definition.

Step to reproduce :
1. Create a new node
2. Fund 3Msat per day
3. Once having ~35M local capacity, the plugin start to open big channels (5M+) with relatively small peers.

ZmnSCPxj:
Fixes: #81

This allows the use-case where the user slowly trickles in money rather than give it a one-time big-time deposit.

### DIFF
--- a/Boss/Mod/ChannelFinderByPopularity.cpp
+++ b/Boss/Mod/ChannelFinderByPopularity.cpp
@@ -5,6 +5,7 @@
 #include"Boss/Msg/CommandRequest.hpp"
 #include"Boss/Msg/CommandResponse.hpp"
 #include"Boss/Msg/Init.hpp"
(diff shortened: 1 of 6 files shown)


ZmnSCPxj:
Good thing I tested this, I caught a bug, ha.

Fixed bug and rebased.

PR #86, opened 2021-12-07
Filter out candidates with low dowsing results

### LINKED ISSUE #85: ChannelCandidatesPreinvestigator to drop Candidates which have too small capacity to their Patron
Following #81 

ChannelCandidatesPreinvestigator do not consider Candidates capacity to their Patron.
As a result, 2/3 of the candidates are useless. They are monitored kept as good candidate until a `ChannelCreator` is called.
ChannelCreator rejected them during channel creation process : 
https://github.com/ZmnSCPxj/clboss/blob/00cdded89cd5eac01ee5c6bb4675ecc7fa20778f/Boss/Mod/ChannelCreator/Planner.cpp#L81-L85

It would be better to directly drop wrong candidates: better visibility, no useless active monitoring, slots not taken by fake candidates, ...

ZmnSCPxj:
Closes: #85

### DIFF
--- a/Boss/Mod/AmountSettingsHandler.cpp
+++ b/Boss/Mod/AmountSettingsHandler.cpp
@@ -1,0 +1,178 @@
+#include"Boss/Mod/AmountSettingsHandler.hpp"
+#include"Boss/Msg/AmountSettings.hpp"
+#include"Boss/Msg/EndOfOptions.hpp"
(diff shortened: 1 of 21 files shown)


ZmnSCPxj:
This interacts with #74.  I might rewrite #74 and put the min/max channel size settings in a separate object that announces the settings, as well as various derived settings, at the start of the plugin.

PR #87, opened 2021-12-09
Centralized size settings

ZmnSCPxj:
Closes: #74

This centralizes the size settings into a single module, which will then emit a message containing the size settings.  As noted in #86, other modules need information like the minimum channel size as well.

@willcl-ark


willcl-ark:
Thanks @ZmnSCPxj and Happy New Year!

I like how this will permit cleaner addition of more options in the future and therefore prefer it to my own approach.

I have taken a cursory look over the changes and these also look good to me (also, it compiles on my machine!), but I still have not tested it. I do notice that it is missing updated documentation, is this something that you you plan to add?

PR #103, opened 2022-04-25
Fix #101

### LINKED ISSUE #101: clboss configs in clconf don't work anymore in 0.11D
I installed 0.11D on my RaspiBlitz recently, used the "clboss-min-onchain=satoshi" comment in the clconf with 0.11B

Now with 0.11D the wallet does not unlock until I comment this out. Same goes for clboss-zerobasefee=.. and clboss-min-channel=...

image

ZmnSCPxj:
Fixes: #101

So, looks like 0.11E is going to release soon....

### DIFF
--- a/Boss/Mod/AmountSettingsHandler.cpp
+++ b/Boss/Mod/AmountSettingsHandler.cpp
@@ -175,4 +175,9 @@ public:
 	      { start(); }
 };
 
+AmountSettingsHandler::~AmountSettingsHandler() =default;
+AmountSettingsHandler::AmountSettingsHandler(AmountSettingsHandler&&) =default;
+
+AmountSettingsHandler::AmountSettingsHandler(S::Bus& bus) : pimpl(Util::make_unique<Impl>(bus)) {}
+
 }}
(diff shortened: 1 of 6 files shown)


robinvonmises:
Sorry to disturb in this case again but how do I fix this issue on my blitz? Wait for 11E? :D

ZmnSCPxj:
Easiest is wait for 0.11E.  Lemme just wait around for a little bit in case other issues come up so there is no need for a 0.11F.

If you are willing to do more hacking on your Pi (and take note that this may void the Raspiblitz warranty, as you would be doing something kinda irregular and I am not a Raspiblitz dev and have no idea how easy it would be) you can run off master.

* Install `build-essential`, `pkg-config`, `libev-dev`, `libcurl4-gnutls-dev`, `libsqlite3-dev`, `dnsutils`, `git`, `automake`, `autoconf-archive`, `libtool`.
* Clone master: `git clone https://github.com/ZmnSCPxj/clboss && cd clboss`.
* `autoreconf -i`
* `./configure && make`
* Modify your Raspiblitz to load CLBOSS from your master, or `sudo make install` if it gets it from `/usr/local/bin` --- this is what will void your Raspiblitz warranty, since you would be poking directly into Raspiblitz.

robinvonmises:
Alright thank you! I will wait then. Although it is difficult to keep my fingers from trying that way, I am not sure in what way this effects the built in node manager script thing, which I have edited to install 0.11D anyway. But thank you any again for that fast reaction and your work ;)

ZmnSCPxj:
https://github.com/ZmnSCPxj/clboss/releases/tag/0.11E

PR #105, opened 2022-04-26
Boss/Mod/PeerCompetitorFeeMonitor/Surveyor.cpp: Use the new `destination` argument to `listchannels` so we do not hammer the RPC too much.

### LINKED ISSUE #100: getting median fees efficiently
I have a strong feeling that I'm missing something, but my C++ skills are horrible, so I'm just gonna ask here:

Boss/Mod/PeerCompetitorFeeMonitor/Surveyor.cpp get's the weighted median for fees for a peer.
it makes one call "listchannels source=PEERID" (line 44) and adds all relevant SCIDs to a stack (line 55).

then it recursively iterates the stack, calling "listchannel scid=SCID) for each element, then figuring out the right direction because it usually get's 2 channels for each SCID, and then it adds the base/proportional fee to some object that eventually gives the weighted median.

as my node has 185 peers with a total of 21634 (x2 directional) channels, this results in 21635 calls to "listchannels scid=SCID" for me. This runs for about 20 minutes.

And I just wonder why not initially make the call "listchannels destination=PEERID" and feed the output directly (minus our own channel) into the median thingy, saving the parsing of 20k channels and 21450 RPC calls? besides switching source/destinatin in the call, it only does one call per peer that is done now, too. I'd guess this should be over in seconds.

ZmnSCPxj:
Fixes: #100

### DIFF
--- a/Boss/Mod/PeerCompetitorFeeMonitor/Main.cpp
+++ b/Boss/Mod/PeerCompetitorFeeMonitor/Main.cpp
@@ -1,5 +1,6 @@
 #include"Boss/Mod/PeerCompetitorFeeMonitor/Main.hpp"
 #include"Boss/Mod/PeerCompetitorFeeMonitor/Surveyor.hpp"
+#include"Boss/Msg/AvailableRpcCommands.hpp"
 #include"Boss/Msg/Init.hpp"
 #include"Boss/Msg/ListpeersAnalyzedResult.hpp"
 #include"Boss/Msg/PeerMedianChannelFee.hpp"
@@ -22,6 +23,9 @@ private:
 	S::Bus& bus;
 	Boss::Mod::Rpc* rpc;
 
+	bool have_listchannels_destination;
+	bool got_available_commands;
+
 	Ln::NodeId self_id;
 
 	/* Current known peers with channels.  */
@@ -35,6 +39,7 @@ public:
 	Impl(S::Bus& bus_
 	    ) : bus(bus_)
 	      , rpc(nullptr)
+	      , got_available_commands(false)
 	      , have_channels(false)
 	      , fired_init(false)
 	      { start(); }
@@ -76,16 +81,50 @@ private:
 				return Ev::lift();
 			return on_periodic();
 		});
+		bus.subscribe<Msg::AvailableRpcCommands
+			     >([this](Msg::AvailableRpcCommands const& m) {
+			got_available_commands = true;
+			auto it = m.commands.find("listchannels");
+			if (it == m.commands.end()) {
+				/* No `listchannels` command...?  */
+				have_listchannels_destination = false;
+				return Ev::lift();
+			}
+			auto& desc = it->second;
+			auto& usage = desc.usage;
+
+			/* Look for [destination] in usage.  */
+			auto static const needle = std::string("[destination]");
+			auto result = std::search( usage.begin(), usage.end()
+						 , needle.begin(), needle.end()
+						 );
+			have_listchannels_destination = (result != usage.end());
+
+			return Boss::log( bus, Debug
+					, "PeerCompetitorFeeMonitor: "
+					  "have_listchannels_destination = %s"
+					, have_listchannels_destination ? "true" : "false"
+					);
+		});
+	}
+
+	Ev::Io<void> wait_available_commands() {
+		return Ev::lift().then([this]() {
+			if (got_available_commands)
+				return Ev::lift();
+			return Ev::yield() + wait_available_commands();
+		});
 	}
 
 	Ev::Io<void> on_periodic() {
-		return Ev::lift().then([this]() {
+		return wait_available_commands().then([this]() {
 			auto f = [this](Ln::NodeId nid) {
 				auto surveyor = Surveyor::create
 						( bus
 						, *rpc
 						, self_id
 						, std::move(nid)
+						, have_listchannels_destination
 						);
 				return surveyor->run();
 			};
--- a/Boss/Mod/PeerCompetitorFeeMonitor/Surveyor.cpp
+++ b/Boss/Mod/PeerCompetitorFeeMonitor/Surveyor.cpp
@@ -36,26 +36,22 @@ Ev::Io<void> Surveyor::core_run() {
 				, std::string(peer_id).c_str()
 				);
 	}).then([this]() {
+		/* Older C-Lightning does not have `destination`
+		 * parameter for `listchannels`.
+		 */
+		auto field = have_listchannels_destination ? "destination" : "source";
 		auto parms = Json::Out()
 			.start_object()
-				.field("source", std::string(peer_id))
+				.field(field, std::string(peer_id))
 			.end_object()
 			;
 		return rpc.command( "listchannels", std::move(parms));
 	}).then([this](Jsmn::Object result) {
 		try {
-			auto cs = result["channels"];
-			for (auto c : cs) {
-				auto dest = Ln::NodeId(std::string(
-					c["destination"]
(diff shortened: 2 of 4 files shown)


ZmnSCPxj:
Warning: not yet tested.

ZmnSCPxj:
Tested on my machine: [clboss-speed-up-surveyor.tar.gz](https://github.com/ZmnSCPxj/clboss/files/8598232/clboss-speed-up-surveyor.tar.gz)

It does reduce the surveyor work tremendously, but note that other modules (such as `FeeModderBySize`) are still hammering `topology`.

ZmnSCPxj:
However, `FeeModderBySize` does limit its hammerring of `topology` once a day (and once at startup), and the other feemodders do not hammer `topology`.  So I guess I will merge this in soon about ~24 hours unless somebody tests the above tarball and complains about something on this PR.

ZmnSCPxj:
Add ChangeLog entry, rebase.

ZmnSCPxj:
Modified to use #115 so that we automatically adapt to the version of the node.

[clboss-speed-up-surveyor-v2.tar.gz](https://github.com/ZmnSCPxj/clboss/files/8609040/clboss-speed-up-surveyor-v2.tar.gz) tested on my machine, also forced `have_listchannels_destination` to `false` and checked too, resulting data was similar but RPC logs show it was indeed not using `destination` in this `false` mode.

I am now looking at `FeeModderBySize` and wondering about the best way to throttle it.

ZmnSCPxj:
Use the idea in https://github.com/ZmnSCPxj/clboss/issues/116#issuecomment-1116328989 to reduce the load `FeeModderBySize` puts on the node software.  Tarball: 
[clboss-speed-up-surveyor-v3.tar.gz](https://github.com/ZmnSCPxj/clboss/files/8618615/clboss-speed-up-surveyor-v3.tar.gz)  Untested.

ZmnSCPxj:
Ouch, on a RaspPi the "get all the channels" technique takes too much time just parsing the result.  Never mind the adding up of the capacities, it takes several minutes just parsing the result, and it jams up the rest of the RPC because parsing takes too long.

The simple parsing code could potentially be optimized (basically, we need a way to guess when one JSON datum ends, we could make a slightly smarter one so that we do only one O(n) pass over the JSON parser instead of ending up with O(n^2)).  But basically the parser is just not responding fast enough to process the large result from a raw `listchannels`.

We could use the same technique Core Lightning does, which is to cheat and look for newlines to separate JSON objects, but that is **NOT** part of standard JSON-RPC 2.0 but an undocumented Core Lightning extension to the protocol, one I am reluctant to lean on.

ZmnSCPxj:
Rebase on master, which now has the optimized JSON parsing code.  Will check on my system and provide a tarball later.

ZmnSCPxj:
Tarball: 
[clboss-speed-up-surveyor-v4.tar.gz](https://github.com/ZmnSCPxj/clboss/files/8637361/clboss-speed-up-surveyor-v4.tar.gz)

@svewa would appreciate if you could check this tarball.  It seems to work reasonably well on my system, though takes 1 minute 30 seconds from `Rpc out: listchannels {}` to `Rpc in: listchannels {} => ...`,  Maybe needs more tweaks for the `Boss::Mod::Rpc` itself.

ZmnSCPxj:
Removed the by-size for now.  Will put that in a separate PR instead.

svewa:
Hi just running v4, can't see the Rpc in/out for not having high enough log level, but from `plugin-clboss: Started.` to `plugin-clboss: FeeModderBySize: Peer 02113fba7b4a54068a335b2a042fd889138f6eb3c791eea6435e144cde90409d47 has 56 other peers, 35 of which have less capacity than us, 21 have more.  Multiplier: 2.60858` it took only 25 seconds. 

Is this O(n) now? parsing the channels once and adding up the satoshis to to the node should be really quick?

ZmnSCPxj:
> Is this O(n) now? parsing the channels once and adding up the satoshis to to the node should be really quick?

Yes, it is now using the method you described.  If you want to participate more in debugging `clboss`, you can set `--log-level=debug:clboss` so that only CLBOSS has debug level, the rest of C-Lightning and other plugins do not have debug log level.

However... Are you running on a RaspPi?  Because on my RaspPi it takes 30 seconds from `Rpc out` to `Rpc in` with a tweaked `Boss::Mod::Rpc` (i.e. *not* the one in the `v4`. tarball).  I need to do additional tweaks on the `Boss::Mod::Rpc`, probably create a different variant of `Jsmn::Parser` --- in particular, `Boss::Mod::Rpc` has its own buffer, and `Jsmn::Parser` has its own buffer, and at least some of the processing is spent moving the same data across buffers.  I will merge this in then open a new PR for the by-size modfiication.

svewa:
Thanks for the tip on logging, will use this to tidy up my logs a lot. Still sad that you can't change the log level at runtime.
I'm on a Intel NUC i5-10, so bit better performance is expected.

svewa:
bit over 4s for me.

```
2022-05-06T10:20:10.146Z DEBUG   plugin-clboss: Rpc out: listchannels {}
2022-05-06T10:20:14.404Z DEBUG   plugin-clboss: Rpc in: listchannels {} => {\"channels\":[{\"source\":\"033533e2db6311c1d417d41279c067d8713ad8a7d577a0d91e65df2f6a82c5b862\",\"destination\":\"039edc94987c8f3adc28dab455efc00dea876089a120f573bd0b0...
```

but I still have 15k calls of `listchannels` with `source` within 45 minutes or so... don't know what for 🤷🏻‍♂️

ZmnSCPxj:
`FeeModderBySize` has to know what nodes your direct peers have peers, since it is comparing you to the peers of your peers.  But that is just one `listchannels` per peer.  Other modules also use `listchannels` --- `fgrep -R "listchannels" Boss/Mod` to see.  `ChannelFinderByDistance` and `ChannelFinderByPopularity` will scan much of the network when they trigger, though those trigger rarely on an old node.  `FundsMover` (which handles any rebalances) will also use `listchannels` to see the short-channel-ids of the peers --- current `listpeers` now gives the `short_channel_id` but I think 0.9.x C-Lightning did not, which is why it uses `listchannels`.  Could switch over to `listpeers` for those but have to do a fallback to `listchannels` if that fails.

Is the node still overloaded from the `listchannels`?  Basically, `FeeModderBySize` is one of the bigger culprits in overloading `listchannels` in parallel --- it does `listchannels` per peer-of-peer, separately, in parallel, which is what is overloading the poor Core Lightning daemon.  Rebalancing just does it twice, one for the source and one for the destination, and it does that sequentially.  `ChannelFinderByDistance` also operates sequentially, waiting for one `listchannels` to complete before going to the next one.  Same with `ChannelFinderByPopularity` --- it does `listchannels` sequentially too.

So if your node memory load on a >100 channels is not too hard, I would rather be lazy and keep the other modules as-is, just modify `FeeModderBySize` by your idea (but requires some modifications on `Boss::Mod::Rpc` to keep it performant on RaspPis), and possibly modify `FudsMover` later to use `listpeers` preferentially (which provides a much smaller result) and fall back to `listchannels` only if the node software version is too old to have `short_channel_id` on the `listpeers` result.

svewa:
Thanks for all the details.

I think the single most important change was the limiting of parallel RPC calls. Since this was implemented I had no problems, just higher system load during times but CLN is always responsive. Of course 99% efficiency gains are always welcome, like you achieved with FeeModderBySize. 

I guess I would not be so tolerant with people running year old CLN but want bleeding edge CLBOSS, but kudos to you for caring about them :)

PR #106, opened 2022-04-26
Boss/Mod/PeerCompetitorFeeMonitor/Main.cpp: Trigger on twice-daily instead of random-hourly.

ZmnSCPxj:
Since C-Lightning itself restricts to only two updates per day anyway,
doing this more often than twice a day is kinda pointless.


ZmnSCPxj:
Against this, we should note that fee changes do get propagted out when a payment fails due to not giving equal or better fees than what we set: insufficient-fee errors include the latest `channel_update`.

PR #107, opened 2022-04-26
Boss/Mod/*: Modify items deprecated in C-Lightning 0.11.

ZmnSCPxj:

PR #115, opened 2022-05-02
Boss/Mod/AvailableRpcCommandsAnnouncer.cpp: New module to announce the commands found on this node.

ZmnSCPxj:
This will allow CLBOSS to figure out what commands (and parameters) the
specific C-Lightning node being managed has, and adapt accordingly.

This is relevant to both #105 (which uses a parameter that is new in 0.10.1) and #107 (so we can adapt to whichever API the node has).


ZmnSCPxj:
[clboss-announce-available-commands.tar.gz](https://github.com/ZmnSCPxj/clboss/files/8602050/clboss-announce-available-commands.tar.gz) for your amusement.

PR #117, opened 2022-05-04
Boss/Mod/Rpc.cpp: Limit max number of concurrent RPC calls.

ZmnSCPxj:
Untested.

### DIFF
--- a/Boss/Mod/Rpc.cpp
+++ b/Boss/Mod/Rpc.cpp
@@ -2,6 +2,7 @@
 #include"Boss/Shutdown.hpp"
 #include"Boss/log.hpp"
 #include"Ev/Io.hpp"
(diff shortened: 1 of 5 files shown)


ZmnSCPxj:
Seems to be working on my machine: 
[clboss-limit-max-rpc.tar.gz](https://github.com/ZmnSCPxj/clboss/files/8617583/clboss-limit-max-rpc.tar.gz)

The above tarball contains only the limiting of the number of concurrent RPCs but stlll uses the old inefficient surveyor code and the current inefficient `FeeModderBySize` code.

Let me let this stew a little bit before I merge in, I worry about memleaks somehow in the new `Ev::Semaphore`.

ZmnSCPxj:
Seems like no leaks so far.  Merging in.

Issue #116, opened 2022-05-02
clboss makes CLN use very very much RAM


svewa:
clboss latest git, but had same issue with 0.11? before.

I run CLN on a node with ~190 peers. I have 20 GB RAM (including 8GB swap)
In normal operation, without clboss, the system uses about 0.8 GB for CLN and a few plugins.

After clboss is finished with `plugin-clboss: PeerCompetitorFeeMonitor: Surveying` I run out of memory. 

`2022-05-02T17:48:10.672Z **BROKEN** 030bec0fb4a5dd5c2dffa5f008210a9b7e0689ee4bf7ae15a7e52491c65fbf4ca5-chan#22195: Could not subdaemon channel: Cannot allocate memory`

`top` shows clightning using all the ram and 100% CPU, sometimes topology also a lot.

ZmnSCPxj:
Probably related to #100 .  Could you try the tarball in #105 and see if that helps?

svewa:
It seems to be the process that kicks in once all peers channels are evaluated. With this tarball the bad behavior starts seconds after I load clboss, and not just after 25 minutes or so with the old code. But it's basically the same. It also blocks CLN, at least I can't `getinfo` or` plugin stop .../clboss` while I watch my RAM being eaten in `top`.

svewa:
With this test also my CLN died. It might be due to clboss, but it also might be for another reason and both just happend when CLN started reacting again. I had no OOM kill and no memory allocation errors in log. Commands were given with tailing the log in the background, which gives some hints about timing here:

`cl@cl:/tmp/clboss-speed-up-surveyor$ cl plugin start /tmp/clboss-speed-up-surveyor/clboss `
```
2022-05-02T23:30:03.126Z INFO    plugin-clboss: clboss 0.11E
2022-05-02T23:30:03.127Z INFO    plugin-clboss: ChannelFeeManager: zerobasefee: allow
2022-05-02T23:30:03.257Z INFO    plugin-clboss: New block at 734636
2022-05-02T23:30:03.258Z INFO    plugin-clboss: DnsSeed: Have 2 seeds.
2022-05-02T23:30:03.258Z INFO    plugin-clboss: Started.
```
[plugin start output with list of plugins etc, successful]
```

2022-05-02T23:30:03.309Z INFO    plugin-clboss: AutoDisconnector: 029542b908ed7e7dbd76f039a53f2614a2e7e9696ef83b5caf872d5aa9ab71e367
[...]

PR #120, opened 2022-05-06
Boss/Mod/FeeModderBySize.cpp: Use a single large query to get all node sizes.

ZmnSCPxj:
Implements: https://github.com/ZmnSCPxj/clboss/issues/116#issuecomment-1116328989

Testing on a RaspPi reveals this takes up a large amount of processing in that environment, so I need to work on getting the basic parts (RPC,  JSON parser) more efficient, possibly by breaking apart the `Jsmn::Parser` abstraction and exposing its internals.


ZmnSCPxj:
Exposed the guts of the JSON parser and have `Boss::Mod::Rpc` write directly into the parser buffer to reduce copying.  Untested.

ZmnSCPxj:
Tarball: 
[clboss-optimize-by-size.tar.gz](https://github.com/ZmnSCPxj/clboss/files/8644378/clboss-optimize-by-size.tar.gz)

Untested.

ZmnSCPxj:
Hmm, still getting a minute or so from `Rpc out: listchannels {}` to `Rpc in: listchannels` on my RaspPi.  Hmm.

ZmnSCPxj:
Okay, so I am getting a minute or so from `Rpc out: listchannels {}` to `Rpc in: listchannels {}` **on startup**, but if I let the node run for a while, it only takes about 14 seconds.  And:

```
$ time sudo -u bitcoin lightning-cli listchannels > /dev/null

real    0m14.690s
user    0m4.610s
sys     0m0.619s
```

So maybe I am at a fairly good optimization point for this.  Probably a bit uncomfortable for smaller nodes running on RaspPis (consider that `JitRebalancer` needs to use the RPC to check if the outgoing channel has capacity, and it blocks all incoming HTLCs until it can do the check) but presumably smaller nodes on RaspPis do not get so many forwards that a 14-second window every about 24 hours is going to do a lot of damage, we hope.  And if you are running a larger node, presumably you are not on a RaspPi.

PR #142, opened 2022-07-14
Boss/Mod/Tally.cpp: Implement `clboss-tally`.

ZmnSCPxj:


ksedgwic:
@chrisguida This looks really useful.  Some thoughts:
- [ ] do we think it is complete? @ZmnSCPxj ?
- [ ] how does it relate to bookkeeper?  Does it have more visibility (or less)?
- [ ] I don't like the `cleartally` operation, would much rather have a way to generate an income/expense report between two specified timestamps (could default to what most people would want).  I think this is trivial to do?

chrisguida:
@ksedgwic yeah, I agree that requesting a report between two timestamps would be amazing to have!

Have you looked at the bkpr commands in CLN yet? Thinking there must be a good way to integrate this CLBOSS data into there. Possibly there are some bkpr account event types that we can just use, if not we can make new types like we did with Smaug.

chrisguida:
I haven't tested this code yet, hopefully I'll get a chance soon!

chrisguida:
Testing this with my mutinynet node running the new signet PR https://github.com/ZmnSCPxj/clboss/pull/207

So far it's opened a couple of channels, nothing in the tally yet:

```
$ lightning-cli clboss-tally
{
   "total": 0,
   "comment": "Reset all tallies to 0 via `clboss-cleartally`.  This data is purely for node operator and CLBOSS will never use this in its heuristics.",
   "history": [
      {
         "time": 1.7165e+09,
         "time_human": "UTC 2024-05-23 20:24:57.352",
         "total": 0
      }
   ]
}
```

chrisguida:
I've rebased this, do you want a new PR @ksedgwic ?

tsjk:
This would be cool to have.

PR #143, opened 2022-07-14
Boss/Mod/EarningsTracker.cpp: Implement `clboss-destroyearningsinfo` command.

### LINKED ISSUE #97: 0.11B not rebalancing because 0.10 spent lots of money rebalancing
I appreciate 0.11B isn't an official release yet, but I really wanted to give it a try as 0.10 seemed to be spending way too much money rebalancing channels. (I picked 0.11B rather than the latest code on master because it had at least been tagged as a candidate release so it felt safer.)

0.11B seems to work well and isn't rebalancing anywhere near as much, but because 0.10 spent so much money rebalancing channels I see messages like this in the log:
```
EarningsRebalancer: Node xxx has bad net income (xxx), refusing to throw good money after bad
```
This might be OK, but I wonder if this means 0.11B is blaming individual nodes for 0.10's prodigality and as a result is going to make these channels less useful than they should be by refusing to rebalance them at all. Should there be some way for 0.11B to ignore money (or at least some money) spent by 0.10 to maintain channels so things start off "fair"?

Is there any way I can reset the "money spent" records for each of my channels to 0? Should I do this even if there is such a way?

(Incidentally, I am very impressed with clboss on the whole. The idea of setting up and managing a lightning node and getting inbound liquidity all seemed like a horrific challenge and it was only the existence of clboss that encouraged me to give it a try at all. A lot of what I'd read made running a node sound a very "social" activity, asking people for inbound liquidity as a kind of favour and risking upsetting them if your node doesn't cooperate well with theirs and so forth. With clboss I don't have to worry about all the warm and fuzzy social relationship stuff. :-) Thanks very much for creating it!)

ZmnSCPxj:
Closes: #97

### DIFF
--- a/Boss/Mod/EarningsTracker.cpp
+++ b/Boss/Mod/EarningsTracker.cpp
@@ -1,6 +1,10 @@
 #include"Boss/Mod/EarningsTracker.hpp"
+#include"Boss/Msg/CommandRequest.hpp"
+#include"Boss/Msg/CommandResponse.hpp"
 #include"Boss/Msg/DbResource.hpp"
 #include"Boss/Msg/ForwardFee.hpp"
+#include"Boss/Msg/Manifestation.hpp"
+#include"Boss/Msg/ManifestCommand.hpp"
 #include"Boss/Msg/ProvideStatus.hpp"
 #include"Boss/Msg/RequestEarningsInfo.hpp"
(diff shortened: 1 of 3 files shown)


ksedgwic:
I think this is superseded by #231 and future work to limit the history range on things which use earnings data.

Issue #148, opened 2022-08-22
Signet support?


matthiasdebernardini:
```
2022-08-22T18:27:42.433Z **BROKEN** plugin-clboss: init: unrecognized network: \"signet\"
2022-08-22T18:27:42.883Z INFO    plugin-clboss: Killing plugin: exited before replying to init
2022-08-22T18:27:42.883Z **BROKEN** plugin-clboss: Plugin marked as important, shutting down lightningd!
2022-08-22T18:27:43.886Z INFO    02ac9367cd4a6484fffc8f458b7e4a043de23af9462a53aaf5f628c9facf436bce-chan#2: Peer transient failure in CHANNELD_NORMAL: channeld: Owning subdaemon channeld died (62208)
```

Does this not work with signet? I am trying to test out clightning in a test enviroment first and thought it would be a good idea to do it with signet.

is there support planned? signet is better for lightning than testnet because the blocks are more consistent

There is already pretty good support for it in the ecosystem (block explorers and wallets). So I think adding it would be a net benefit.

PR #149, opened 2022-08-22
Adding Signet option

matthiasdebernardini:
This is referencing the issue #148 adds Signet support by adding the keywords.

It seems to work but I am not sure how to guarantee quality, if there are tests then I would be happy to run them.

As far as I can tell, the only thing missing is being able to test boltz integration 
https://github.com/ZmnSCPxj/clboss/blob/3379983b5a849caa46bb405d07c134e843c248c9/Boss/Mod/BoltzSwapper/Main.cpp#L28

### DIFF
--- a/Boss/Mod/Initiator.cpp
+++ b/Boss/Mod/Initiator.cpp
@@ -188,6 +188,8 @@ public:
 				network = Boss::Msg::Network_Bitcoin;
 			else if (network_s == "testnet")
 				network = Boss::Msg::Network_Testnet;
+			else if (network_s == "signet")
+				network = Boss::Msg::Network_Signet;
 			else if (network_s == "regtest")
 				network = Boss::Msg::Network_Regtest;
 			else
--- a/Boss/Msg/Network.hpp
+++ b/Boss/Msg/Network.hpp
@@ -7,6 +7,7 @@ enum Network
 { Network_Bitcoin
 , Network_Testnet
 , Network_Regtest
+, Network_Signet
 };
 
 } }


openoms:
Signet support would be welcome as it is preferred over testnet3.

ksedgwic:
mempool.space thinks signet has 35 lightning nodes while testnet shows 470 and mainnet has 15,487

I'm pretty sure signet would require a patch to min_node_to_process see ZmnSCPxj/clboss@ad49445

I suspect this parameter should be different for each network

ksedgwic:
I added a PR which sets `min_nodes_to_process` to a network specific value: https://github.com/ZmnSCPxj/clboss/pull/173

vincenzopalazzo (commented):
OK make sense, if you help to mantaining it.

Also signet is a little bit strange because can be more signet network.

clboss keep a list of nodes (aka seeds) to start connecting with, so this is PR is missing to add it. In addition, we should make sure that this peer should be living enough.

An alternative approach is specify at least one seed as plugin option, this could help with signet, but not sure how the clboss metrics will react with a single node

chrisguida:
Ended up reimplementing this myself before I knew this PR existed https://github.com/ZmnSCPxj/clboss/pull/207

xD

Let's just merge this, it works great for now

chrisguida (approved):
Works with mutinynet

ksedgwic:
Looks good

vincenzopalazzo (commented):
Sorry for being late to the discussion, but I think this PR is missing a list of seed nodes for signet. Or am I missing something?

Additionally, we may want to provide a seed list via a command line option, considering it is unclear how many signet network there are.

chrisguida:
Ah I forgot about seed nodes. You can just connect to the faucet node on mutinynet, you can find the URI on https://faucet.mutinynet.com

ksedgwic:
There is another PR for updating the mainnet seed nodes.  The CLBOSS support isn't very good on any network.  I generally make a few manual seed connections and bootstrap gossip that way ...

Maybe we should make a DNS based thing so at least the nodes aren't compiled in ...

ksedgwic:
to clarify, I only make initial gossip connections manually, everything else works automatically

Issue #147, opened 2022-08-14
Assertion `min_amount + min_remaining <= max_amount' failed


JaviLib:
Using master or 0.13A on arm64 I get this error when running the plugin right at the start:

```
Boss/Mod/ChannelCreator/Planner.cpp:55: Boss::Mod::ChannelCreator::Planner::Impl::Impl(Boss::Mod::ChannelCreator::Planner::DowserFunc, Ln::Amount, std::vector<std::pair<Ln::NodeId, Ln::NodeId> >, std::size_t, Ln::Amount, Ln::Amount, Ln::Amount): Assertion `min_amount + min_remaining <= max_amount' failed.
```

JaviLib:
It continues to happen, but only when I setup `clboss-max-channel` or `clboss-min-channel`.

ksedgwic:
I'm not sure, but this seems unlikely to be caused by the `msat` conversion because it predates it, I believe.

I haven't run w/ `clboss-max-channel` or `clboss-min-channel` yet but they seem useful to our case.  I will try them when I get a chance and look for that error ...

tsjk:
I haven't managed to get those flags to work, ever. They make clboss crash. At least the max one. This occurred before the msat removal as well...

JaviLib:
It seems that it works at some values. After long testing the options, I found that the only ones that work for me are:

```
important-plugin=/usr/local/bin/clboss
clboss-auto-close=true
#clboss-min-channel=500000
clboss-max-channel=2000000
clboss-zerobasefee=require
clboss-min-onchain=0
``` 

As you see, I had to comment out `clboss-min-channel`, put `clboss-max-channel` higher or equal than 2M, and configure `clboss-min-onchain` to zero. I haven't tested `clboss-max-channel` higher than 2M, but it is surely crashing with lower values.

ksedgwic:
Is this the same problem as #137

tsjk:
Most likely, yes. Both regard the setting of the limits.

tsjk:
Actually, no. The other has the assert on the line above failing.

tsjk:
This was discussed yesterday on the discord. According to
https://github.com/ZmnSCPxj/clboss/blob/b6795c43dc646225902845c81c4a4df3fea532e7/Boss/Mod/AmountSettingsHandler.cpp#L153-L158
and
[...]

PR #162, opened 2023-05-03
Update to handle JSON-RPC amounts in either the old or new conventions

ksedgwic:
The JSON-RPC interface has changed the type of amounts from string w/ "msat" suffix to a plain number.

The backtrace hack and removal of the Amount constructor-from-string are useful during debugging but we might want to drop them ...

### DIFF
--- a/Boss/Mod/ActiveProber.cpp
+++ b/Boss/Mod/ActiveProber.cpp
@@ -137,9 +137,9 @@ private:
 						chan0 = Ln::Scid(std::string(
 							c["short_channel_id"]
 						));
-						cap0 = Ln::Amount(std::string(
+						cap0 = Ln::Amount::object(
 							c["spendable_msat"]
-						));
+						);
 						break;
 					}
 
@@ -263,9 +263,9 @@ private:
 				chan1 = Ln::Scid(std::string(
 					hop1["channel"]
 				));
-				amount1 = Ln::Amount(std::string(
+				amount1 = Ln::Amount::object(
 					hop1["amount_msat"]
-				));
+				);
 				delay1 = std::uint32_t(double(
 					hop1["delay"]
 				));
--- a/Boss/Mod/ChannelFundsComputer.cpp
+++ b/Boss/Mod/ChannelFundsComputer.cpp
@@ -23,12 +23,9 @@ ChannelFundsComputer::on_listfunds(Msg::ListfundsResult const& r) {
 			if (!chan.is_object() || !chan.has("our_amount_msat"))
 				continue;
 			auto amount_j = chan["our_amount_msat"];
-			if (!amount_j.is_string())
+			if (!Ln::Amount::valid_object(amount_j))
 				continue;
-			auto amount_s = std::string(amount_j);
-			if (!Ln::Amount::valid_string(amount_s))
-				continue;
-			auto amount = Ln::Amount(amount_s);
+			auto amount = Ln::Amount::object(amount_j);
 			total += amount;
 
 			if (!chan.has("connected"))
--- a/Boss/Mod/Dowser.cpp
+++ b/Boss/Mod/Dowser.cpp
@@ -243,10 +243,9 @@ private:
 				   )
 					continue;
 				auto amt_j = chan["amount_msat"];
-				if (!amt_j.is_string())
+				if (!Ln::Amount::valid_object(amt_j))
 					continue;
-				auto amt_s = std::string(amt_j);
-				return Ev::lift(Ln::Amount(amt_s));
+				return Ev::lift(Ln::Amount::object(amt_j));
 			}
 
 			/* The channel *can* disappear between the time
--- a/Boss/Mod/EarningsRebalancer.cpp
+++ b/Boss/Mod/EarningsRebalancer.cpp
@@ -173,12 +173,12 @@ private:
 					   )
 						continue;
 
-					auto to_us = Ln::Amount(std::string(
+					auto to_us = Ln::Amount::object(
 						c["to_us_msat"]
-					));
-					auto c_total = Ln::Amount(std::string(
+					);
+					auto c_total = Ln::Amount::object(
 						c["total_msat"]
-					));
+					);
 					auto to_them = c_total - to_us;
 
 					spendable += to_us;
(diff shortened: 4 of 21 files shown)


tsjk:
What does this fix?
Does it make clboss compatible with cln v23?

ksedgwic:
Yes, this PR makes CLBOSS compatible with `v23.05`.

However, it is not currently backwards compatible; if you run this PR against older versions of CLN it won't work.  It should be possible to make it backwards compatible, I will have a look at that next ...

ksedgwic:
@tsjk I've reworked the CLBOSS amount PR so it should be able to handle either the old or new conventions.

ksedgwic:
It's running successfully on CLN v23.05, but I have not verified it on an older CLN.

tsjk:
I'll try to find time to test this soon.

ksedgwic:
@tsjk switching back to Draft, I'm seeing bogus swap reports in `clboss-status`:
```
         {
            "time": 1.68391e+09,
            "time_human": "UTC 2023-05-12 17:29:15.552",
            "amount_sent": "2msat",
            "amount_received": "18446744073709551456msat",
            "amount_lost": "0msat",
            "provider": "Boltz::Service(\"https://testnet.boltz.exchange/api\")"
         },
```
The `amount_sent` and `amount_received` are obviously wrong.

The values in the `data.clboss` db are also bad after my changes:
```
1679187607.98172|299528669|297715000|Boltz::Service("https://testnet.boltz.exchange/api")
1679200208.08367|100094092|99148000|Boltz::Service("https://testnet.boltz.exchange/api")
1679211008.25419|398096693|395157000|Boltz::Service("https://testnet.boltz.exchange/api")
1679291804.75702|108338105|105921000|Boltz::Service("https://testnet.boltz.exchange/api")
1679302004.89865|100000095|92722000|Boltz::Service("https://testnet.boltz.exchange/api")
1680386607.52125|4096999375|4077496000|Boltz::Service("https://testnet.boltz.exchange/api")
1681898663.87814|312002872|309660000|Boltz::Service("https://testnet.boltz.exchange/api")
1683855569.11341|2|94067341782384|Boltz::Service("https://testnet.boltz.exchange/api")
1683867701.28066|2|94842369897008|Boltz::Service("https://testnet.boltz.exchange/api")
1683878085.26563|2|-160|Boltz::Service("https://testnet.boltz.exchange/api")
1683885285.61652|2|-160|Boltz::Service("https://testnet.boltz.exchange/api")
1683893817.04882|2|-160|Boltz::Service("https://testnet.boltz.exchange/api")
1683902355.21184|2|-160|Boltz::Service("https://testnet.boltz.exchange/api")
1683912555.55297|2|-160|Boltz::Service("https://testnet.boltz.exchange/api")
```

Still looking, clues welcome ...

tsjk:
The negative numbers don't look too good. :)
I have not detailed knowledge of the code, but I have browsed through it on several occasions.
The amount received is on-chain, right? Are on-chain transactions effected in the same way by the api-change in cln?
If you compare what the numbers should be with what they are - does that give clues? How much did you actually send and receive?

ksedgwic:
@tsjk I force pushed a new version which uses a class method instead of a conversion constructor for the JSON -> Amount method.  I have seen one swap since and it was correctly reported.  Possibly the problem is fixed.

TODO:
- [x] observe additional swap reports with v23.05
- [ ] test on older versions which still use deprecated amount msat string format.

ksedgwic:
@tsjk I've observed more swaps with well-formed reports.  I'm checking the first TODO box.  Any chance you could try on an older CLN version to validate backwards compatibility?

tsjk:
Sorry. I've been busy with other things.
I guess I could test this. Haven't ran cln on a testnet but can imagine to try it and let clboss go wild.
Not sure testing it on my mainnet node would make a lot of sense.

JaviLib:
When is the solution going to be merged with master? Is it ok to wait?

ksedgwic:
It still needs testing on older systems with the legacy msat format, modulo that it has been working great for us on `v23.05`

tsjk:
I ended up upgrading my mainnet node, because the version I was using (v22.11.1 with some back-ported patches) actually back-ported the msat removal - voiding my reason for stay on that version. For now my build recipe for Gentoo only applies this patch if the installed version of core lightning is v23.05 or higher.
But, we should definitely try to test this. My idea was to setup at testnet node with a testnet bitcoind. I realise that's quite an effort. So, perhaps this should be tested by someone still on earlier version - or we could just go high chaparral and requiring that the latest clboss requires core-lightning v23.05 or later...

ksedgwic:
> or we could just go high chaparral and requiring that the latest clboss requires core-lightning v23.05 or later...

Over time this will become more and more reasonable.  Not sure if it is quite reasonable yet though (especially for mainnet where folks run more mature versions ...)

JaviLib:
Well, people using old versions of Core Lightning and not intending to upgrade have zero reasons to try this patch. So I think it is better to just release this and pass an important notice like "Not compatible with versions previous to v23.05". You may want to even test version in code and return an error if version is not equal or higher than v23.05..

tsjk:
Yes, checking for version sounds sensible.
On another note, I've been running this for a while and it seems ok. No submarine swaps performed as of yet, however.

JaviLib:
Would you please check if #147 is related to this?

chrisguida:
Testing this on 23.02 with a raspiblitz

vincenzopalazzo (approved):
ACK https://github.com/ZmnSCPxj/clboss/pull/162/commits/a4bb0192550803db3d07628a29284a76f7204365

chrisguida:
Update: ran overnight on raspiblitz with CLN 23.02, looks great. Let's merge!

ACK a4bb0192550803db3d07628a29284a76f7204365

chrisguida (approved):

ksedgwic:
I'm attempting resolve a conflict with recent commits on `master`.

Having a build problem with master before any rebasing/merging: #166

ksedgwic:
rebasing on master

ksedgwic:
I'm going to smoke test this in my test nodes, will merge if problem free

tsjk:
Was just about to ask whether this was tested on any pre-msat-drop nodes in the end.

ksedgwic:
@tsjk https://github.com/ZmnSCPxj/clboss/pull/162#issuecomment-1682803646

ksedgwic:
Looks stable on my testnet nodes, merging

kn0wmad:
Any ETA on a release that might include this fix?

ksedgwic:
@kn0wmad https://github.com/ZmnSCPxj/clboss/issues/172

PR #171, opened 2023-08-18
Fix rpc error: blockheight: should be an integer

ksedgwic:
Fix required when the number of blocks is large enough to trigger a floating represention, then it fails to parse


vincenzopalazzo (approved):
LGMT

Currently running on my node

ksedgwic:
I've been running with this for a long time

PR #173, opened 2023-08-21
network: Make the min_nodes_to_process network specific

ksedgwic:
TODO:
- [x] Validate network specific behavior
- [x] Address signet if the default of 10 is not appropriate

### DIFF
--- a/Boss/Mod/ChannelFinderByPopularity.cpp
+++ b/Boss/Mod/ChannelFinderByPopularity.cpp
@@ -54,14 +54,6 @@ auto const max_proposals = size_t(5);
  */
 auto const min_channels_to_backoff = size_t(4);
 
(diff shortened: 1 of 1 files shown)


ksedgwic:
@vincenzopalazzo Does this PR look ok to you?  This allows CLBOSS to "engage" on `testnet`.  I'm not exactly sure how many nodes to expect when folks run in `regtest` or `signet` but lowering the threshold to 10 seems likely to be better than the current `mainnet` setting of 800 ...

vincenzopalazzo (approved):
utACK https://github.com/ZmnSCPxj/clboss/pull/173/commits/c1bb9c6b0b80e8528f1b9c0f90fd3575d725ee2e

Make sense to me.

Sorry I miss the notification of this PR

PR #207, opened 2024-05-23
add signet

### LINKED ISSUE #148: Signet support?

chrisguida:
Add support for networks named `signet`. Tested only with mutinynet so far. Seems to be opening channels just fine, but obviously swaps won't work until Boltz supports mutinynet/signet swaps (or we add a swapping method other than Boltz).

Closes #148

### DIFF
--- a/Boss/Mod/Initiator.cpp
+++ b/Boss/Mod/Initiator.cpp
@@ -189,6 +189,8 @@ public:
 				network = Boss::Msg::Network_Bitcoin;
 			else if (network_s == "testnet")
 				network = Boss::Msg::Network_Testnet;
+			else if (network_s == "signet")
+				network = Boss::Msg::Network_Signet;
 			else if (network_s == "regtest")
 				network = Boss::Msg::Network_Regtest;
 			else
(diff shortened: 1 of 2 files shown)


chrisguida:
Darn

`plugin-clboss: ChannelFinderByPopularity: Not enough nodes in routemap (only 186, need 300), will try again later.`

Reverting that line...

chrisguida:
Ok this PR is now exactly identical to #149 , let's just merge that instead. Closing.

PR #209, opened 2024-06-01
testnet: reduce the min_nodes_to_process because testnet is shrinking

ksedgwic:
Changelog-Changed: testnet: The minimum number of nodes (with channels) before the ChannelFinderByPopularity starts is reduced from 300 to 200.


vincenzopalazzo (approved):
ACK https://github.com/ZmnSCPxj/clboss/pull/209/commits/8c07aa6f8721bd30f8e6e25e51db9c24158e0541

Issue #202, opened 2024-05-07
Running clboss in a "low cost" mode


sj-fisher:
Is there currently any support for running clboss in a "low cost" mode?

I have a node which I use to make and receive fairly small and infrequent payments over lightning. I would like to run clboss on this node in order to benefit from:
- having the channel fees automatically managed
- having the success_per_* statistics maintained for each channel
- having candidate nodes for new channels evaluated

I am not trying to make a profit from routing or to help the network by operating a routing node at a loss, so I do not want clboss to automatically do anything which costs money:
- automatic channel opening
- rebalancing via Boltz
- rebalancing via payments to self
- automatic channel closing (which can already be disabled)

It would be nice to have the ability to manually re-enable those features temporarily if I decide the node would benefit from them.

My current half-baked thinking is that the way to do this would be by adding a handful of options/new behaviors to clboss which might be individually useful, such as:
- only allow automatic channel opening if explicitly enabled, and for a time-limited period only
- only allow rebalancing via Boltz if explicitly enabled, and for a time-limited period only
- only allow rebalancing via payments to self if explicitly enabled, and for a time-limited period only

Combining these would then effectively provide a variety of "economy modes" for clboss with an emphasis on lowering costs in return for the node being less reliable and hands-off.

Noting the obvious pattern in the ideas above and that you can already do the opposite with "noticing on-chain funds" - you can temporarily disable this for a set period - I wonder if the thing to do here would be to generalize that pattern. In other words, add some generic way to configure a clboss feature to be on or off by default and to have an ability to temporarily invert that setting for a time-limited period.

I am interested in doing some implementation work to support this, if it isn't already a thing. I've opened this issue before I start hacking to see if anyone has any thoughts. Is this even a sensible thing to want? Is there an alternative automated Core Lightning node manager which suits these goals better than clboss? Is there a better way to modify clboss to achieve this goal instead of the approach I sketched out above?

ksedgwic:
Wow, I just added https://github.com/ZmnSCPxj/clboss/issues/201 yesterday!  I think it's mostly the same sentiment ...

> Is this even a sensible thing to want?

Yes, this seems very sensible.

ksedgwic:
One tricky part is figuring out how the various CLBOSS modules interact.  For example many of the features use the `EarningsTracker` so we'll need to ensure that disabling feature `A` doesn't impede feature `B` ... 

I've started to make diagrams to improve my own understanding:
[...]

PR #212, opened 2024-06-05
Add options to disable clboss features which cost money

sj-fisher:
This attempts to address https://github.com/ZmnSCPxj/clboss/issues/202.

This is a draft pull request - feedback would be appreciated!

This has been lightly tested on signet and nothing more. Please be very cautious if using this on mainnet!

# Overview

This adds three new boolean options:
- clboss-auto-open-channels - is clboss allowed to open channels?
- clboss-auto-swap - is clboss allowed to perform offchain-to-onchain swaps?
- clboss-circular-rebalance - is clboss allowed to perform circular rebalancing?

The new options can be specified via config file entries, e.g.
```
important-plugin=/path/to/clboss
clboss-auto-open-channels=false
```

and can also be controlled via lightning-cli, e.g.:
```
$ lightning-cli clboss-auto-open-channels false
```

If present, config file entries are applied on startup, overriding any changes made earlier using lightning-cli commands. Otherwise, changes made with lightning-cli are persisted in the database across restarts. This feels a bit over-complex, but it's also the most flexible approach.

If nothing is ever specified, all three default to "true" for backwards compatibility.

Each of the options can also be temporarily set using lightning-cli, e.g.:
```
$ lightning-cli clboss-temporarily-auto-open-channels true 3
```
will temporarily allow auto-opening of channels for 3 hours. If this command uses the same boolean value as the base state, it is a no-op.

A temporary setting can be removed before it times out by using the non-temporary form of the command. For example, to cancel the previous temporary permission to auto-open channels:
```
$ lightning-cli clboss-auto-open-channels false
```

You can see the current state using:
```
$ lightning-cli clboss-status
```

# Possible problems

## Overlap between auto-open-channels and ignore-onchain

The new auto-open-channels/temporarily-auto-open-channels commands have a lot of overlap with the existing notice/ignore-onchain commands. They are not quite the same:
- auto-open-channels can default to "false" and be temporarily set to "true", whereas ignore-onchain is always temporary and the base state is always "notice"
- they are checked in different places in the code (ignore-onchain in OnchainFundsAnnouncer::on_block(), auto-open-channels in ChannelCreationDecider::Impl::run())

It would obviously be possible to check in the same places in the code if desired. I do like the ability to have "default false", which is why I added this in the first place. I think there is a theoretical difference between the two which doesn't apply at the moment - if (for example) we had support for advertising our willingness to dual-fund channels, ignore-onchain should probably prevent that, whereas "auto-open-channels false" would not.

I'm reluctant to break backwards compatibility by removing the notice/ignore-onchain options completely, but they could be handled as special-case syntax for the new auto-open-channels behavior.

# Implementation notes

Most of the new code is in Boss/ModG/TimedBoolean.hpp, which adds a TimedBoolean helper class which deals with all the messy user interface and state maintenance for "timed booleans", i.e. booleans which have a base state and which can be set to a different state for a fixed period of time. This has its origins in the existing OnchainFundsIgnorer code.

Three new controllers (AutoOpenController, AutoSwapController and CircularRebalanceController, all in Boss/Mod) use TimedBoolean internally and expose a pair of Request/Response messages to allow their current state to be queried.

Logic has then been added (in ChannelCreationDecider, NodeBalanceSwapper and EarningsRebalancer) to disable the existing behavior where the relevant timed boolean is currently false.


ksedgwic:
Thanks for working on this!

I kicked off the CI, I think there might be some interesting failures.  Some thoughts on other parts follow ...

ksedgwic:
When looking at `AutoOpenController`, `AutoSwapController`, and `CircularBalanceController` side-by-side I'm struck w/ how similar the `.hpp` and `.cpp` files are.  Is there a way to use templates (or other techniques) to reduce the code duplication?

(I'm tempted to see what ChatGPT proposes ... ;-) )

ksedgwic:
You've modified the `EarningsRebalancer` to inhibit rebalancing, but what about the `InitialRebalancer` and the `JitRebalancer`?  Do they need similar treatment?

See: https://github.com/ksedgwic/clboss/blob/df51d5486b05acbeca16cb9e42d18fa082e6079a/docs/channel_balancing.md

I wonder if it would make more sense to inhibit the actions of the `FundsMover` instead?  That would cover all three rebalancers ...

I haven't looked into what interactions the `FundsMover` has outside channel balancing, important to make sure it isn't required for other things ...

ksedgwic:
I think the `NodeBalanceSwapper` might have a similar issue, I believe the `NeedsOnchainFundsSwapper` would also require similar inhibition.  Can we inhibit the more general `SwapManager` or `BoltzSwapper` instead?

https://github.com/ksedgwic/clboss/blob/17c56feafa8dbe7ff872e280aebeaa20ad18cb95/docs/offchain_to_onchain_swap.md

ksedgwic:
We should make sure that inhibiting the `ChannelCreationDecider` is enough:

https://github.com/ksedgwic/clboss/blob/17c56feafa8dbe7ff872e280aebeaa20ad18cb95/docs/channel_creation.md

Your implementation here looks more likely to cut off all the desired activity.

sj-fisher:
Thanks Ken. I will have a look at the test failures (I must admit I overlooked the test suite completely) and see what I can do about the other points you've raised. Thanks very much for taking a look! If you do feel interested to pursue the code duplication thing with ChatGPT assistance obviously feel free to go ahead with that.

I didn't touch JitRebalancer because I think it's disabled by default, although it may well be good to handle this case too. InitialRebalancer seems to be limited to spending 0.04% of the channel balance and is a one-off thing so I thought it might be best to allow that. Your suggestion of moving the check into FundsMover (subject to verifying it doesn't break anything else) seems superficially reasonable, though I haven't looked into this in any detail yet.

I will be away a bit over the next week or so and while I might be able to work on this, I might not, so apologies in advance if I'm a bit unresponsive.

sj-fisher:
Incidentally is there any way to convert these diagrams like channel_creation.md into a high-resolution bitmap file for viewing offline? I have experimented a bit and I haven't been able to convert the whole diagram into a bitmap using the upstream website - I could only get a bitmap showing what was currently on my screen. I find panning and zooming around them pretty uncomfortable in github and in the upstream website viewer and they would probably be a lot more usable if I could use a local graphics viewer on them.

ksedgwic:
Yeah, the github viewer is not great.  The best I can do with the github viewer is:
1. Open the page enclosing the diagram full-screen -then-
2. Click the `<->` icon in the upper right corner to bring up a light box

A much better viewer that might be able to export is here: https://mermaid.live/

I'd love to figure out how to have a link to click to preload the CLBOSS graphs into that editor but haven't figured it out yet ...

ksedgwic:
An example w/ one of the CLBOSS graphs:
https://mermaid-js.github.io/mermaid-live-editor/view#pako:eNq1Vttu2zAM_RVDRYANSIvcL34Y0LTrVqDNuqRPg4FBtelEiCNpstw0jfPvoy9JfGuTAd2TLeociiIPaW-ILRwgJjEMo1bbMM60aWws4npiZc-p0haJ1w64NPD0BLgDClRktQh4C4tst8a2VrO4xdHFTFE5Nx5H6fIW3TGqhTo__xJGi3AMK-o46jvljoduItB0ReU95XQGMWwCfwLwdQqsIuRNEYen4Ku7cYzAZ2Q-3cEEfCm4D7tTMzGlV_H12gNjJDz9Gm1KUIbLPM88G7pPdV8rsQDzrN1up-_nK-boudmRLzF75Al78aiovUjOi9dh1ltVKqbCYzbTkfVnIDTVTPCjpEvbBnmMkzVEpAclnpkDeVYuCRWk6PVKwWlgF0BRDX6xRri9WJUDTHcVEvJ7mWJkjtzV4sk9XoucKK-fsPQiUDaUbvDIlqD8CBS_NRv3jAfRDYrAO-ZrSdd-hZ52W5UcN-COf8m5CLid0PZW5GO3lZNaKaMiaor1C_wrsVxiQFkdxfYyPK-fB7q-5c-CYUbSJ1rebdX9JQuJqPSeCi0NJRtpZQdEEhNL6YEGJ15NQAqlC0oYAzj-D44Di_GbKH-FFj1FFldzyjl4O0Ffg41xxkHsHAOEbx10xEOJ9r6jtzYjVxbRis0wPb99NH76bJGwqjfGONVH1KOorP-Wi9IRh27AZi8rO7amyn6DXDb_w52z8viogXCQ22mNlcOjkJmLDZLcYrc6zMscOHU7ARvYc8LYm-I2q4gnYzi1ubL4XUQpJjzEWwnfh5PMtrAQcakW8uPKII8OQHl6jXbQbIfdzrhQ-dn9DXSKSDZvPDoL3_tax-lKc1PEydzErHRdEU6RfkqBZTaag3D2E5_UCX7NlpQ5-Ne3iWjYXnNYQvJPl_7oWcTiW4TSQIvpmtvE1CqAOgmkg1_xa0bxT2-ZN351GBaPmC71fDRKyn8Jscfgkpgb8kLMZmNw0e33h4Nmr9HttTrtYZ2sidlqtC8azU67NWj1e53usNXd1slr7KFxMeg0-r1mv9_uDhA_6NQJxKfdJz-vtuAum5HtXymI6wk

sj-fisher:
Apologies for the delay in working on this, Ken, and thanks again for taking the time to give such detailed comments. I've now been over them in more detail.

Thanks also for the advice about viewing the charts in mermaid.live, that's definitely more usable.

> You've modified the `EarningsRebalancer` to inhibit rebalancing, but what about the `InitialRebalancer` and the `JitRebalancer`? Do they need similar treatment?
> 
> See: https://github.com/ksedgwic/clboss/blob/df51d5486b05acbeca16cb9e42d18fa082e6079a/docs/channel_balancing.md
> 
> I wonder if it would make more sense to inhibit the actions of the `FundsMover` instead? That would cover all three rebalancers ...
> 
> I haven't looked into what interactions the `FundsMover` has outside channel balancing, important to make sure it isn't required for other things ...

That diagram suggests that ActiveProber uses FundsMover, which doesn't seem obvious to me when looking at the code. Am I getting confused? It looks to me as though ActiveProber::sendpay() makes a direct "sendpay" RPC.

FundsMover appears to be invoked in general by Msg::RequestMoveFunds, which is raised by MoveFundsCommand, InitialRebalancer, JitRebalancer and EarningsRebalancer. So I think the main concern with moving the "inhibit circular rebalance" behavior into FundsMover would be the possibility of breaking MoveFundsCommand. That is debug-only, so maybe it doesn't really matter. Otherwise, I think it would work fairly well to move the inhibit logic into FundsMover/Main.cpp's Msg::RequestMoveFunds subscription handler.

I was worried that inhibiting rebalances might cause problems where we think we've sent a payment and it never arrives and we blame one of our peers, but I think that's fine. This appears to be handled by monitoring "sendpay" RPC calls, so if we simply never make a "sendpay" call in the first place there is no risk of anyone getting blamed.

> I think the `NodeBalanceSwapper` might have a similar issue, I believe the `NeedsOnchainFundsSwapper` would also require similar inhibition. Can we inhibit the more general `SwapManager` or `BoltzSwapper` instead?
> 
> https://github.com/ksedgwic/clboss/blob/17c56feafa8dbe7ff872e280aebeaa20ad18cb95/docs/offchain_to_onchain_swap.md

I think you're right about NeedsOnchainFundsSwapper needing to be included in the restriction.

I'm not so keen on inhibiting in SwapManager or BoltzSwapper. For BoltzSwapper specifically, as this is a general "no auto swap" restriction it doesn't seem to belong in a Boltz-specific class. SwapManager feels (admittedly based on my imperfect understanding of the clboss architecture) more like code to do a job than code to implement policy like this.

My inclination would be to move this check from NodeBalanceSwapper into trigger_swap(), which is used by both NodeBalanceSwapper and NeedsOnchainFundsSwapper (and triggers SwapManager). 

Incidentally, trigger_swap() is one of the two places the existing GetOnchainIgnoreFlag is checked, which surprised me at first but does make sense the more I think about it - there is no point triggering a swap to get on-chain funds if we will just refuse to do anything with them. (The other place is OnChainFundsAnnouncer.) I mention this mainly as supporting evidence that trigger_swap() is a reasonable place to add "policy" code like this inhibit behavior.

> We should make sure that inhibiting the `ChannelCreationDecider` is enough:
> 
> https://github.com/ksedgwic/clboss/blob/17c56feafa8dbe7ff872e280aebeaa20ad18cb95/docs/channel_creation.md
> 
> Your implementation here looks more likely to cut off all the desired activity.

I've had a look over the code and your diagram and it looks to me like ChannelCreationDecider is the right place for this logic - it is the only code that can raise Msg::RequestChannelCreation messages. If I've missed something here could you please let me know what?

> When looking at `AutoOpenController`, `AutoSwapController`, and `CircularBalanceController` side-by-side I'm struck w/ how similar the `.hpp` and `.cpp` files are. Is there a way to use templates (or other techniques) to reduce the code duplication?
> 
> (I'm tempted to see what ChatGPT proposes ... ;-) )

I haven't had a look at this yet, although I agree it would be good to reduce the code duplication. I will see if I can think of anything - I'll hold off making any changes based on the above analysis for at least a couple of days in case you or anyone else has any comments.

ksedgwic:
> That diagram suggests that ActiveProber uses FundsMover, which doesn't seem obvious to me when looking at the code. Am I getting confused? It looks to me as though ActiveProber::sendpay() makes a direct "sendpay" RPC.

I think `ActiveProber` sends `SolicitDeletablePaymentLabelFilter` to `FundsMover`.  I think it is not requesting that funds be moved but rather the `ActiveProber` is trying to identify previously moved funds as "probed" (instead of moved for other purposes).

ksedgwic:
> So I think the main concern with moving the "inhibit circular rebalance" behavior into FundsMover would be the possibility of breaking MoveFundsCommand. That is debug-only, so maybe it doesn't really matter. Otherwise, I think it would work fairly well to move the inhibit logic into FundsMover/Main.cpp's Msg::RequestMoveFunds subscription handler.

Ah, I missed the `MoveFundsCommand` pushed a diagram fix PR: https://github.com/ZmnSCPxj/clboss/pull/220

Thank you!

sj-fisher:
> > That diagram suggests that ActiveProber uses FundsMover, which doesn't seem obvious to me when looking at the code. Am I getting confused? It looks to me as though ActiveProber::sendpay() makes a direct "sendpay" RPC.
> 
> I think `ActiveProber` sends `SolicitDeletablePaymentLabelFilter` to `FundsMover`. I think it is not requesting that funds be moved but rather the `ActiveProber` is trying to identify previously moved funds as "probed" (instead of moved for other purposes).

Thanks Ken, I think I missed the label on the arrow somehow. I agree this is not using the FundsMover to move anything.

That said, I may be getting confused but it looks to me as though both ActiveProber and FundsMover *subscribe* to the SolicitDeletablePaymentLabelFilter message but it is only raised in PaymentDeleter::Impl::initialize(), so ActiveProber doesn't use FundsMover via this message, it is just that they both respond to this message to allow PaymentDeleter to delete their payments.

sj-fisher:
> When looking at `AutoOpenController`, `AutoSwapController`, and `CircularBalanceController` side-by-side I'm struck w/ how similar the `.hpp` and `.cpp` files are. Is there a way to use templates (or other techniques) to reduce the code duplication?

I have pushed a commit to the low-cost-v1 branch which tries to address this. It's a fairly obvious template-based approach but while slightly fiddly (because you can't use literal strings as template arguments) it does save code duplication.

sj-fisher:
I've pushed another commit to low-cost-v1 which moves the swap and circular rebalance inhibit logic as discussed above. This has had minimal testing on signet only.

I believe this addresses all the feedback so far. It definitely needs some testing before it can be considered for merging, and of course if anyone has any further feedback please let me know.

ksedgwic:
@sj-fisher are you on discord?  Possibly related thoughts here: https://discord.com/channels/899980449231814676/907338137041256509/1273205821098037279

sj-fisher:
I just signed up for discord but when I click that link I get a page which says "No text channels". Do I need to join a particular server or channel first? I'm not very familiar with how this works, but I am able to browse some random discoverable servers and see messages.

ksedgwic:
That link still works for me ... maybe you need an invite?  Try this:
https://discord.gg/RJhszzc6

The comment is in the CLBOSS channel on the core lightning discord server, all very public ...

sj-fisher:
Thanks Ken, I can see it now I've accepted your invitation. Maybe this was necessary or maybe I just missed something before.

I am not yet running 0.13.3 so I have not seen this fundsmover activity or how/if it interacts with the changes on this pull request. Now I know about 0.13.3 I will look at upgrading, but it's not likely to be for a few weeks as I want to make sure I have spare time to try to recover things if it all goes bad. I'll also try to keep an eye on the discord now I know about it, thanks.

ksedgwic:
@sj-fisher we're trying to setup a conf call to brainstorm CLBOSS futures:
https://discord.com/channels/899980449231814676/907338137041256509/1282346104360534148

sj-fisher:
Thanks Ken, I sent you a message on discord about the call.

On a different note and speaking to anyone reading, I have created a branch https://github.com/sj-fisher/clboss/tree/low-cost-v1-rebased which contains the same changes as low-cost-v1 but on top of clboss 0.13.3. I have given this a very quick test on signet and will probably try it on my mainnet node soon-ish. Testing welcome but please use caution if you're on mainnet!

chrisguida:
Should I test this?

sj-fisher:
Apologies for not replying earlier...

It may be best to see what Ken thinks. I implemented this because I had a small node I wanted to run clboss on but didn't want it splurging on Boltz fees to get inbound liquidity I mostly didn't care about.

That said, I completely get (what I think is) Ken's position from the recent Friday calls that it's better to tune clboss to work without manual options like this as far as possible. Given that, I can't see this being merged into a release any time soon - it would probably only get merged after the ongoing tuning and discoverability work nears completion and if it turns out there is still value in these options.

I don't think these changes are likely to break things badly, so if you want these options to work around clboss wasting money right now, please do go ahead and test. But if you're just testing to help these changes moving towards inclusion, that would be appreciated and a step forward but may not actually result in them moving much closer to inclusion as I suspect the tuning/discoverability work is the main blocker.

(By "discoverability" I mean the development of tools and wiki pages and so on which allow users to have a better idea what's going on inside their node.)

sj-fisher:
If you do decide to test, please note the new branch https://github.com/sj-fisher/clboss/tree/low-cost-v1-rebased where I rebased this change onto 0.13.3. It wasn't a difficult merge so you'd probably be fine taking the low-cost-v1 branch mentioned on the actual pull request, but you might as well take the one where I already did it.

ksedgwic:
#263 partially addresses this, maybe

sj-fisher:
Thanks Ken. I agree it looks like there is some overlap with #263.

I have just rebased my change - not doing anything to take advantage of that possible overlap to simplify things - onto master's ec4d69579653d454317c9b9eb704aa1775aab256 commit. The merge was trivial and I have given it extremely minimal testing on signet so far. The code is on https://github.com/sj-fisher/clboss/tree/low-cost-v1-rebased-2 if anyone wants it.

sj-fisher:
I've rebased my changes onto a recent master on branch https://github.com/sj-fisher/clboss/tree/low-cost-v1-rebase-3 This has had minimal testing on my own node. I skimmed the diff and it doesn't look like there was any major conflict.

Issue #185, opened 2023-12-06
Crash on startup when --developer used


ksedgwic:
version:
`lightningd: v23.11-14-g2f5f098-modded` (VLS build on `v23.11`)

relevant config:
```
log-level=info
clboss-auto-close=true
important-plugin=/usr/local/bin/clboss
developer
```

result:
```
Dec 06 13:38:25 home4 lightningd[669956]: 2023-12-06T21:38:24.681Z INFO    lightningd: v23.11-14-g2f5f098-modded
Dec 06 13:38:25 home4 lightningd[669956]: 2023-12-06T21:38:25.278Z DEBUG   lightningd: Opened log file -
Dec 06 13:38:25 home4 lightningd[669956]: 2023-12-06T21:38:25.278Z DEBUG   lightningd: Opened log file /home/cln/.lightning/testnet/lightning.log
Dec 06 13:38:25 home4 lightningd[669956]: 2023-12-06T21:38:25.305Z DEBUG   plugin-manager: started(669961) /usr/local/libexec/c-lightning/plugins/autoclean
Dec 06 13:38:25 home4 lightningd[669956]: 2023-12-06T21:38:25.345Z DEBUG   plugin-manager: started(669962) /usr/local/libexec/c-lightning/plugins/chanbackup
Dec 06 13:38:25 home4 lightningd[669956]: 2023-12-06T21:38:25.370Z DEBUG   plugin-manager: started(669963) /usr/local/libexec/c-lightning/plugins/bcli
Dec 06 13:38:25 home4 lightningd[669956]: 2023-12-06T21:38:25.395Z DEBUG   plugin-manager: started(669964) /usr/local/libexec/c-lightning/plugins/commando
Dec 06 13:38:25 home4 lightningd[669956]: 2023-12-06T21:38:25.422Z DEBUG   plugin-manager: started(669968) /usr/local/libexec/c-lightning/plugins/fetchinvoice
Dec 06 13:38:25 home4 lightningd[669956]: 2023-12-06T21:38:25.447Z DEBUG   plugin-manager: started(669969) /usr/local/libexec/c-lightning/plugins/funder
Dec 06 13:38:25 home4 lightningd[669956]: 2023-12-06T21:38:25.471Z DEBUG   plugin-manager: started(669970) /usr/local/libexec/c-lightning/plugins/topology
[...]

PR #216, opened 2024-06-15
Recognize `--developer` CLI flag and don't exit giving usage ([#185])

ksedgwic:
CLBOSS did not know the `--developer` flag and generated a usage message and exited.  Fixed with:
- recognize and consume the `developer` flag, no actions for now
- print unrecognized options to std::cerr

Fixes ([#185])

### DIFF
--- a/Boss/Main.cpp
+++ b/Boss/Main.cpp
@@ -74,8 +74,16 @@ public:
 				if (system(os.str().c_str()))
 					;
 				kill(getpid(), SIGSTOP);
-			} else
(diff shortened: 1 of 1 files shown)


vincenzopalazzo (approved):
ACK https://github.com/ZmnSCPxj/clboss/pull/216/commits/e839c2a065e634459cd5c78815c699a6850826e5

Thanks for doing it

farscapian:
I can confirm that this change has resolved the issue I was seeing with LNPlay, which relies on the --developer flag.

PR #219, opened 2024-06-26
Avoid vector out of bounds access in ParserExposedBuffer.cpp

vasild:
Accessing elements past the size of a `std::vector` is undefined behavior [1] and is actually checked in recent versions of LLVM libcxx, which is used in FreeBSD [2].

[1] https://en.cppreference.com/w/cpp/container/vector/operator_at:
> Accessing a nonexistent element through this operator is undefined
> behavior.

[2] https://cgit.freebsd.org/src/tree/contrib/llvm-project/libcxx/include/vector?h=2472e352d80fcf6440fd42fbb16960cc49d05b03#n1393

### DIFF
--- a/Jsmn/ParserExposedBuffer.cpp
+++ b/Jsmn/ParserExposedBuffer.cpp
@@ -105,6 +105,9 @@ private:
 				return nullptr;
 		}
 		for (;;) {
+			if (start_idx == end_idx) {
+				return nullptr;
+			}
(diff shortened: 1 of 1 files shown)


ksedgwic:
@vasild Thanks for submitting ... will look

vincenzopalazzo (approved):
LGTM, thanks!

Issue #222, opened 2024-07-21
EarningsTracker not seeing fees


ksedgwic:
My node did some forwarding and collected some fees, from `summars`:
```
                                         forwards                                         
 resolved_time         in_channel        out_channel       in_sats   out_sats   fee_msats 
 7/20/24, 6:39:40 PM   HOPPINGSQUIRREL   LNBϟG [Hub-3]     164,790    164,784       6,505 
 7/20/24, 6:39:45 PM   HOPPINGSQUIRREL   LNBϟG [Hub-3]     678,118    678,092      26,524 
 7/20/24, 6:39:49 PM   HOPPINGSQUIRREL   LNBϟG [Hub-3]     157,736    157,730       6,230 
 7/20/24, 6:39:50 PM   HOPPINGSQUIRREL   LNBϟG [Hub-3]     407,088    407,072      15,954 
 7/20/24, 6:40:20 PM   HOPPINGSQUIRREL   LNBϟG [Hub-2]     137,818    137,813       5,374 
 7/20/24, 6:40:42 PM   HOPPINGSQUIRREL   LNBϟG [Hub-2]     109,292    109,288       4,262 
 7/20/24, 6:40:45 PM   HOPPINGSQUIRREL   LNBϟG [Hub-2]     118,154    118,149       4,607 
```

But the `offchain_earnings_tracker` output does not show any fees:
```
   "offchain_earnings_tracker": {
      "029fe435040c8b665f731f2b0c81d039238ef1e3a1b1de0afac2b476361a26d675": {
         "in_earnings": 0,
         "in_expenditures": 0,
         "out_earnings": 0,
         "out_expenditures": 0
      },
      "02c91d6aa51aa940608b497b6beebcb1aec05be3c47704b682b3889424679ca490": {
         "in_earnings": 0,
         "in_expenditures": 0,
         "out_earnings": 0,
         "out_expenditures": 0
      },
```

Inspecting the `ForwardFeeMonitor` code shows a likely culprit:
```
			auto payload = n.params["forward_event"];
			if ( !payload.has("out_channel")
			  || !payload.has("fee")          <= used to be sent as compat for `fee_msat`
			  || !payload.has("resolved_time")
			  || !payload.has("received_time")
			   )
				return Ev::lift();
			if (std::string(payload["status"]) != "settled")
				return Ev::lift();
```

In CLN `v23.05` the deprecated compatibility non-msat values were removed:
[...]

PR #223, opened 2024-07-21
Restore ForwardFeeMonitor's ability to see forwarding fees

ksedgwic:
Fixes ([#222])
    
Prior to ElementsProject/lightning@780f32d (`v23.05`) both `fee` and
`fee_msat` were sent for compatibility.  The ForwardFeeMonitor was
checking for the presence of the `fee` field before processing the
record.  This needed to be updated to `fee_msat`.


ksedgwic:
See also #227

PR #225, opened 2024-07-28
Add Util::BacktraceException wrapper

ksedgwic:
## Abstract

Added `Util::BacktraceException` which captures backtraces where an exception is thrown and then formats them for debugging when they are displayed with `what()`.  The backtraces are more useful if the following configuration is used: `./configure CXXFLAGS="-g -Og"` but this results in larger, less optimized binaries.

## Motivation

When an unexpected exception occurs on a user's machine we typically get a very terse "parse error" or similar message.  This PR adds logging of the stack backtrace where the exception was encountered, greatly increasing the likelihood of finding the problem from the reported log alone.

### DIFF
--- a/Bitcoin/Tx.cpp
+++ b/Bitcoin/Tx.cpp
@@ -88,9 +88,9 @@ Tx::Tx(std::string const& s) {
 	auto is = std::istringstream(std::move(str));
 	is >> *this;
 	if (!is.good())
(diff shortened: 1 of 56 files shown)


vincenzopalazzo (approved):
ACK https://github.com/ZmnSCPxj/clboss/pull/225/commits/7e4f07093ee8fb6a83854662793c14c7eb249091

PR #226, opened 2024-07-31
contrib: Add clboss-channel-stats to summarize fee income performance

ksedgwic:
This script produces a summary of channel routing stats, sorted by performance:

- Displays PeerID, SCID, and Alias for each channel
- Sorts first by net fees (income - expenses) (descending)
- Then sorts by "Success Per Day" (SPD) (descending)
- Then sorts by Age in days (ascending)

The channels at the top of the list are good, the ones at the bottom are bad.

### DIFF
--- a/contrib/clboss-routing-stats
+++ b/contrib/clboss-routing-stats
@@ -1,0 +1,154 @@
+#!/usr/bin/env python3
+
+# This script produces a summary of channel stats, sorted by routing performance:
+#
+# - Displays PeerID, SCID, and Alias for each channel
+# - Sorts first by net fees (income - expenses) (descending)
+# - Then sorts by "Success Per Day" (SPD) (descending)
+# - Then sorts by Age in days (ascending)
+#
+# The channels at the top of the list are good, the ones at the bottom are bad.
+
+import subprocess
+import argparse
+import json
+from tabulate import tabulate
+from wcwidth import wcswidth
+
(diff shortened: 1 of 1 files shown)


vincenzopalazzo (commented):
LGTM, do you think that this is too hard to include inside CLBOSS?

ksedgwic:
That was my first approach but it is quite a bit of code and would risk
CLBOSS stability itself. Putting it outside is very straightforward and I
think much less risky.

I think we should experiment with it outside first

--
Ken Sedgwick
Bonsai Software, Inc.
http://www.bonsai.com/ken/
(510) 269-7334
***@***.***
Public Key: http://www.bonsai.com/ken/ken.asc
GPG Fingerprint: 4695 E5B8 F781 BF85 4326  9639 BBFC E515 8602 5550

contrib/clboss-routing-stats
```diff
@@ -0,0 +1,150 @@
...
+    if not listpeerchannels_data:
+        return
+
+    channels_data = listpeerchannels_data.get("channels", [])
+
+    channels = {}
+    peers = {}
+    for channel in channels_data:
+        # Skip channels that are closed
+        if channel.get("state") == "ONCHAIN":
```
tsjk: I'd skip all states except CHANNELD_NORMAL.

ksedgwic: agreed, done

Issue #227, opened 2024-08-02
Upgrade EarningsTracker to make date-based income/expense available


ksedgwic:
### Current Earnings Tracking:

Currently fee income from routing and expense due to rebalancing are held in the `EarningsTracker` table on a per-peer basis:
```sql
			CREATE TABLE IF NOT EXISTS "EarningsTracker"
			     ( node TEXT PRIMARY KEY
			     , in_earnings INTEGER NOT NULL
			     , in_expenditures INTEGER NOT NULL
			     , out_earnings INTEGER NOT NULL
			     , out_expenditures INTEGER NOT NULL
			     );
```

Problems:

#### 1. Ancient data can dilute/overwhelm current data

Long running channels have income/expense data dominated by old accumulated data.  Since channels and peer environments change over time this may not provide the best information for current decisions.
Related:
- https://github.com/ZmnSCPxj/clboss/issues/97
- https://github.com/ZmnSCPxj/clboss/pull/143/files

#### 2. Users would like income/expense reports for specific periods of time

Would help answer questions like:
- How is my node changing over time?
- What was the effect of the change I made last week?
Related:
- https://github.com/ZmnSCPxj/clboss/pull/142

#### 3. Doesn't "heal"

This is sort of a special case of #1 ... if a bug (ie #222 ) causes distortions (either actual or measured) it would be really nice if the instance "healed" over time and converged on sensible decisions ...

### Possible solutions:
1. Expunge old data from the `EarningsTracker` table.  I hate removing data ...
2. Make all earnings and expenditures timestamped.  Makes table really big, queries take longer ...
3. Time buckets

### Time buckets

Add a `time_bucket` field to `EarningsTracker`:
```sql
CREATE TABLE IF NOT EXISTS "EarningsTracker"
(
    node TEXT,
    time_bucket INTEGER,
    in_earnings INTEGER NOT NULL,
    in_expenditures INTEGER NOT NULL,
    out_earnings INTEGER NOT NULL,
    out_expenditures INTEGER NOT NULL,
    PRIMARY KEY (node, time_bucket)
);
```

When earnings or expenditures are being reported, the current unix timestamp is divided by the number of seconds in a day to derive a "time bucket".

When a query is made for income/expense it can specify a desired time bucket range (or all if the current behavior is desired).

We can consider expunging/archiving old data in the future.

ksedgwic:
Closing as fixed

PR #228, opened 2024-08-06
Add time_buckets to EarningsTracker data

ksedgwic:
Addresses #227 

**IMPORTANT - This PR makes a schema change and migrates data!**

## Abstract

This PR modifies the `EarningsTracker` schema to hold data in "buckets", one per day.  This PR only modifies the collection and storage,  the strategies and existing status continue to use earnings data from all time for now.

## Motivation

As described in #227 there are several shortcomings w/ the current storage of `EarningsTracker` data:
1. Ancient data can dilute/overwhelm current data
2. Users would like income/expense reports for specific periods of time (especially recent periods)
3. The data doesn't "heal" as bugs are fixed and strategies are improved.

This PR enables time-based data earnings collection, future PR's should investigate modifying the strategies and reports to take advantage of the ability to filter by time range.

## Specification

This PR updates the `EarningsTracker` table to have a "time bucket" column.  The timestamps of incoming fee and balance events are quantized to a time bucket, currently one day.

The existing strategies and reports use the sql `SUM` to return the same answers that they would with the current schema.

The time and nodeid columns are indexed for efficient operations.

```
			CREATE TABLE IF NOT EXISTS EarningsTracker
			     ( node TEXT NOT NULL
			     , time_bucket REAL NOT NULL
			     , in_earnings INTEGER NOT NULL
			     , in_expenditures INTEGER NOT NULL
			     , out_earnings INTEGER NOT NULL
			     , out_expenditures INTEGER NOT NULL
			     , PRIMARY KEY (node, time_bucket)
			     );
			CREATE INDEX IF NOT EXISTS
			    idx_earnings_tracker_node_time ON EarningsTracker (node, time_bucket);
			CREATE INDEX IF NOT EXISTS
			    idx_earnings_tracker_time_node ON EarningsTracker (time_bucket, node);
```

Existing legacy data is migrated into the table with a time bucket value of 0.  This allows it to be considered and used but eventually be aged out.

Two new CLI commands are added to allow inspecting the new data:
- **`clboss-recent-earnings`**:
  - **Purpose**: Returns a data structure equivalent to the
    `offchain_earnings_tracker` collection in `clboss-status`, but
    only includes recent earnings and expenditures.
  - **Arguments**:
    - `days` (optional): Specifies the number of days to include in
      the report. Defaults to a fortnight (14 days) if not provided.

- **`clboss-earnings-history`**:
  - **Purpose**: Provides a daily breakdown of earnings and expenditures.
  - **Arguments**:
    - `nodeid` (optional): Limits the history to a particular node if
      provided. Without this argument, the values are accumulated
      across all peers.
  - **Output**: 
    - The history consists of an array of records showing the earnings
      and expenditures for each day.
    - The history includes an initial record with a time value of 0,
      which contains any legacy earnings and expenditures collected by
      CLBOSS before daily tracking was implemented.

### DIFF
--- a/Boss/Mod/EarningsTracker.cpp
+++ b/Boss/Mod/EarningsTracker.cpp
@@ -12,6 +12,8 @@
 #include"S/Bus.hpp"
 #include"Sqlite3.hpp"
 #include"Util/make_unique.hpp"
+
+#include<cmath>
 #include<map>
 
 namespace Boss { namespace Mod {
@@ -294,7 +296,13 @@ public:
 EarningsTracker::EarningsTracker(EarningsTracker&&) =default;
 EarningsTracker::~EarningsTracker() =default;
 
-		EarningsTracker::EarningsTracker(S::Bus& bus, std::function<double()> get_now_ )
+EarningsTracker::EarningsTracker(S::Bus& bus, std::function<double()> get_now_ )
 	: pimpl(Util::make_unique<Impl>(bus, get_now_)) { }
+
+double EarningsTracker::bucket_time(double input_time) {
+	constexpr double seconds_per_day = 24 * 60 * 60;
+	double bucket_start = std::floor(input_time / seconds_per_day) * seconds_per_day;
+	return bucket_start;
+}
 
 }}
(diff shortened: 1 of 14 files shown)


tsjk:
From the users perspective - is anything needed to be done when upgrading to a version that includes this? Or is it "just" that once the upgrade is carried out, downgrading requires some extra manual labor.

ksedgwic:
Nothing needs to be done by the user for forward migration.  Downgrading is possible (w/o loss of any data) but would need to be done manually or by future code.

ksedgwic:
This is subsumed by #230

PR #239, opened 2024-10-02
ensure that commit_hash.h is included in the source tarball

ksedgwic:


ksedgwic:
This is not needed anymore (already incorporated)

PR #240, opened 2024-10-02
add explicit github release action

ksedgwic:
This was necessary because the default tarball preparation did not include commit_hash.h


ksedgwic:
No longer relevant, release action is added separately

PR #242, opened 2024-10-09
Avoid program_invocation_name which is available only on Linux

vasild:
`program_invocation_name` is defined by the system headers only on Linux, or at least it is not available on FreeBSD.

Thus, maintain own global variable that contains the program name.

### DIFF
--- a/Boss/Main.cpp
+++ b/Boss/Main.cpp
@@ -20,6 +20,8 @@
 # include "config.h"
 #endif
 
+std::string g_argv0;
+
 namespace Boss {
 
 class Main::Impl {
@@ -62,6 +64,7 @@ public:
 	      {
 		assert(argv.size() >= 1);
 		argv0 = argv[0];
+		g_argv0 = argv[0];
 		if (argv.size() >= 2) {
 			auto argv1 = argv[1];
 			if (argv1 == "--version" || argv1 == "-V")
--- a/Util/BacktraceException.hpp
+++ b/Util/BacktraceException.hpp
@@ -6,8 +6,15 @@
 #include <iomanip>
 #include <memory>
 #include <sstream>
-#include <vector>
+#include <stdlib.h>
 #include <string.h>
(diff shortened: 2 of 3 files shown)


Boss/Main.cpp
```diff
@@ -62,6 +64,7 @@ class Main::Impl {
 	      {
 		assert(argv.size() >= 1);
 		argv0 = argv[0];
+		g_argv0 = argv0.c_str();
```
vasild: Does this have a lifetime issue? If `Main::Impl::argv0` is destroyed, then the global `g_argv0` will become dangling. Maybe copy the string to a global buffer?

ksedgwic: Yes, I think that is dangerous.  ChatGPT recommends a fix:
https://chatgpt.com/share/6712ab8d-b780-8008-8bba-468bea99aad2

Can you fix and re-push?

vasild: Done, used `std::string` (which holds the data) rather than a bare pointer, pointing to some other place that can be destroyed too early.

ksedgwic:
I found a problem w/ the unit tests; they don't set the g_argv0 var on startup.  Normally they don't have unhandled exceptions, but when developing they do and are very useful.

I have a patch to use the built-ins as a fallback.  I'll try to push it here ...

ksedgwic:
@vasild could you make sure that compiles with __FreeBSD__?

vasild:
Yes, it compiles and I squashed the two commits and made some further changes:

* test for the presence of `getprogname()` instead of assuming "if FreeBSD then `getprogname()` must be present" because that would exclude other platforms that have `getprogname()` but are not FreeBSD.
* return `std::string` from the new `progname()` to avoid potential issues with the lifetime - returning a pointer to something which if changed invalidates the pointer. Now the `.c_str()` usage is limited to the call of `snprintf()`.

PR #243, opened 2024-10-09
contrib: add --lightning-dir option

endothermicdev:
This adds a custom lightning directory option to clboss-earnings-history, clboss-recent-earnings, and clboss-routing-stats for users with non-default configurations.  I probably should have just symlinked it, but I think it can be handy to have these options be explicit.


endothermicdev:
By the way, great stats summary tools Ken!

vincenzopalazzo (approved):

PR #256, opened 2024-12-04
Add --disable-exception-backtrace configure option

ksedgwic:
Because some platforms have issues w/ exception backtraces


chrisguida:
Can confirm that `./configure --disable-exception-backtrace` before building has fixed both of my broken machines

ACK https://github.com/ZmnSCPxj/clboss/pull/256/commits/80534951e5c4476acc237d2d91420617cf48d446

chrisguida (approved):

PR #262, opened 2025-06-04
Add sys_stats_report to generate heartbeat status reports w/ nostr

ksedgwic:
Add an example script that can be used w/ https://github.com/ksedgwic/privdm to send periodic status reports as nostr private DMs.

This script needs to be adjusted for paths and other system dependent things.

### DIFF
--- a/contrib/sys_stats_report
+++ b/contrib/sys_stats_report
@@ -2,6 +2,13 @@
 
 export NO_COLOR=1
 
+# Optional --testnet flag propagates to all lightning-cli commands.
+LIGHTNING_CLI="lightning-cli"
+if [[ "$1" == "--testnet" ]]; then
+  LIGHTNING_CLI="lightning-cli --testnet"
+  shift
+fi
+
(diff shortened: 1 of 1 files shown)


chrisguida:
Hmm, curious if you have any idea what this error means?

```
bitcoin@rpb-um350:~/cl-plugins-available/clboss/contrib$ ./sys_stats_report 

 00:16:01 up 1 day, 22:54,  3 users,  load average: 1.11, 1.26, 1.34
Unit cln-mainnet.service could not be found.
Cannot open /var/log/sysstat/sa08: No such file or directory
Please check if data collecting is enabled


awk: cmd. line:1: BEGIN{printf "%0.0f", *1e11}
awk: cmd. line:1:                       ^ syntax error
awk: cmd. line:1: BEGIN{printf "%0.0f", *1e11}
awk: cmd. line:1:                       ^ syntax error
awk: cmd. line:2: (FILENAME=- FNR=1) fatal: attempt to access field -1
Our balance at start of bucket: 0 msat
feerates: -1 -1 -1 -1 unknown


```

PR #263, opened 2025-06-24
Add configurable max rebalance fee in ppm

ksedgwic:
Introduce `--clboss-max-rebalance-fee-ppm` to cap the fee allowed for a single rebalance.

Both JitRebalancer and EarningsRebalancer register and use this option, defaulting to 5000 ppm (0.5%). 

Documentation updated to explain the new setting


ksedgwic:
Currently the same config parameter sets the max fee ppm for both the `EarningsRebalancer` and the `JitRebalancer` ... would we ever want to set them differently?  (It's easy to do)

PR #265, opened 2025-07-02
Add clboss-feerates RPC

ksedgwic:
Add clboss-feerates so we don't have to run clboss-status just to get the fee rates ...
    
    - Register and implement the new clboss-feerates command in OnchainFeeMonitor
    - Document the command in README and note it in the changelog
    - Update RPC manifest tests to include the new command


tsjk:
Reviewed the code and tested on CLN v25.05. LGTM.

PR #271, opened 2025-09-19
contrib: resolve python warning in earnings-history

endothermicdev:
When running under python 3.12.3, clboss-earnings-history produces the warning: clboss-earnings-history:36: DeprecationWarning: datetime.datetime.utcfromtimestamp() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.fromtimestamp(timestamp, datetime.UTC).

### DIFF
--- a/contrib/clboss-earnings-history
+++ b/contrib/clboss-earnings-history
@@ -5,7 +5,12 @@ import csv
 import json
 import os
 import subprocess
-from datetime import datetime, UTC
+try:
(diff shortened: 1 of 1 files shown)


ksedgwic:
@endothermicdev I pushed a commit to this PR to allow it work on old python versions too.  Can you check and make sure it still works on your machine?

endothermicdev:
Still works!

PR #277, opened 2025-11-30
docs: fixed duplicate sections for earnings reports

smolting:
There were near-identical sections for the earnings reports. The one with the [TBD] placeholder text was removed.


ksedgwic:
@smolting Thanks for fixing this!

PR #292, opened 2026-01-12
Add "all" (by_node) mode to clboss-earnings-history API

ksedgwic:
Needed for the aggregate (percentile) view of earnings history #291 

Previously the `clboss-earnings-history` API either returned:
1. node-id specific data by day (when node_id is specified)
2. the sum of earnings for all nodes for each day (when node_id is not specified)

This PR adds a special node_id arg "all" which reports data for all node_ids but preserves the per-node records ...

- Adds "all" (by_node) mode to clboss-earnings-history: nodeid == "all" returns per-node daily buckets (preserving individual node records) instead of collapsing earnings into a single summed daily value.
- Implementation: introduces a by_node flag threaded into earnings_history_report (signature updated), adjusts SQL to select/order by time_bucket and node when by_node is true, and includes a node field in each JSON history entry; also enforces that by_node requires an empty node_id.
- Docs & tests: README updated to document the "all" behavior and a new test validates per-node history output (127 entries).

### DIFF
--- a/Boss/Mod/EarningsTracker.cpp
+++ b/Boss/Mod/EarningsTracker.cpp
@@ -614,10 +614,8 @@ private:
 		                   out_expenditures,
 		                   out_rebalanced
 		              FROM "EarningsTracker"
+			     ORDER BY time_bucket, node;
 		        )QRY";
-			if (!nodeid.empty())
-				sql += " WHERE node = :nodeid";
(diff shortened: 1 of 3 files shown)


ksedgwic:
rebased onto master

ksedgwic:
merging, this is a prerequisite for the fee modder plotting tools
tested on my production node

PR #297, opened 2026-02-06
Harden Parse Failure Handling in Boltz code

ksedgwic:
- don't crash
- log the error diagnostics

- Hardens JSON parse failure handling in NormalConnection.cpp by tracking parse errors, stopping processing after failures, and raising diagnostic "Invalid JSON" errors instead of allowing crashes to terminate execution

- Adds error-tolerant feerate parsing in Env.cpp with fallback values and logging for parsing failures, replacing the previous direct extraction approach that assumed presence and type of rate values

- Wraps API calls to convert unexpected exceptions into Boltz::ApiError, normalizing exception handling while preserving intended Boltz::ApiError flows


ksedgwic:
Superseded by #302

PR #301, opened 2026-02-13
fix: use CLN decode RPC in InvoicePayer, add unit test

### LINKED ISSUE #299: CLN 26.04 will disable 'decodepay' RPC call
'decodepay' was deprecated a very, very long time ago. CLN 24.04 will remove the call, the all purpose 'decode' must be used instead. The fix will be simple, it is hard to imagine that anyone is still using CLN versions several years old on mainnet.

ksedgwic:
Fixes https://github.com/ksedgwic/clboss/issues/299

- Replaces the deprecated "decodepay" RPC with "decode" in InvoicePayer.cpp, including fixing the RPC argument name to "string" (was "bolt11") and adding post-response validation to ensure a "bolt11 invoice" type with valid=true; error messaging updated accordingly.
- Adds a comprehensive unit test tests/boss/test_invoicepayer_decodepay.cpp that exercises the decode+pay flow end-to-end using a mocked JSON-RPC server, test doubles, and async messaging.
- Updates README.md and the test harness (Makefile.am) to reflect the change from "decodepay" to "decode".

### DIFF
--- a/Boss/Mod/InvoicePayer.cpp
+++ b/Boss/Mod/InvoicePayer.cpp
@@ -72,8 +72,17 @@ Ev::Io<void> InvoicePayer::pay(std::string n_invoice) {
 				.field("bolt11", *inv)
 			.end_object()
 			;
-		return rpc->command("decodepay", std::move(parms));
+		return rpc->command("decode", std::move(parms));
 	}).then([this, inv](Jsmn::Object res) {
+		if (!res.has("type")
+		|| std::string(res["type"]) != "bolt11 invoice"
+		|| !res.has("valid")
+		|| !res["valid"].is_boolean()
+		|| !bool(res["valid"])
+		) {
+			throw Jsmn::TypeError();
+		}
+
 		/* Check the features and see if this is MPP-enabled.  */
 		auto is_mpp = false;
 		if (res.has("features")) {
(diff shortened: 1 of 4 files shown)


ksedgwic:
@tank-welder this look ok to you?

tank-welder:
The `decode` RPC call should fail with this change.

on line 72 (Boss/Mod/InvoicePayer.cpp), you supply an argument with the key "bolt11", but that argument name does not exist for the "decode" RPC call. "string" is the named arg that "decode" requires.

I will take the unit test for a spin and report back.

ksedgwic:
I think you are correct, note the unit test is self-fulfilling and doesn't test this aspect (correctly).

ksedgwic:
I pushed a fix (and fixed the unit test too)

I'll go test this in situ ... to make sure

ksedgwic:
I confirmed that CLN wants the named arg to be "string":
```
lightning-cli --testnet -k decode bolt11="$INVOICE"
{
   "code": -32602,
   "message": "unknown parameter: bolt11, this may be caused by a failure to autodetect key=value-style parameters. Please try using the -k flag and explicit key=value pairs of parameters."
}

lightning-cli --testnet -k decode string="$INVOICE"
<happy decode output>
```

I think this is now correct.  Thank you @tank-welder for the issue and review attention!

PR #305, opened 2026-03-01
Fix: Replace std::result_of with std::invoke_result_t for C++20

lduchosal:
std::result_of was deprecated in C++17 and removed in C++20. This patch replaces it with std::invoke_result_t which is the modern equivalent.

Fixes build on systems with C++20 compilers (e.g., FreeBSD 14+ with clang 18).

Replaces deprecated `std::result_of` with `std::invoke_result_t` in Ev/Io.hpp and Ev/map.hpp to ensure C++20 compatibility, addressing removal of `std::result_of` in C++20 standard.

Updates two `then()` template methods in Ev::Io class and the public `map()` function to use `std::invoke_result_t` for computing return types, maintaining equivalent functionality while using the modern type trait.

Minimal changes across two headers (4 lines modified total) with no impact to control flow or execution logic—only type computation improvements.

### DIFF
--- a/Ev/Io.hpp
+++ b/Ev/Io.hpp
@@ -91,9 +91,9 @@ public:
 
 	/* (>>=) :: IO a -> (a -> IO b) -> IO b*/
 	template<typename f>
-	Io<typename Detail::IoInner<typename std::result_of<f(a)>::type>::type>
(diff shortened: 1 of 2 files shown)


ksedgwic:
@lduchosal thank you for this PR!  I think you saved us from `clang` breakage on this release.

PR #313, opened 2026-05-19
Adapt to CLN v26.06 deprecation surface

ksedgwic:
CLN v26.06 deprecates getroute (replaced by the askrene-backed
getroutes) and removes the "unknown" sentinel value from
channel_state_changed.old_state (the field is now omitted
instead).  On lightningd configurations where deprecated APIs
are disabled -- developer mode or explicit
allow-deprecated-apis=false -- the deprecation surface becomes
JSON-RPC -32601 hard errors, and CLBOSS effectively stops
working:

- Every Dowser probe fails, so ChannelCandidateInvestigator's
  Janitor rejects every candidate at 0msat and no channels get
  opened.
- Every Matchmaker getroute call fails, so candidate proposals
  that need a patron found are dropped on the floor.
- Every ActiveProber getroute call fails, so no peer
  uptime/capacity probes happen.
- Every channel_state_changed notification produces a noisy
  "Unexpected channel_state_changed payload" Error log line.

This PR migrates three of CLBOSS's four getroute call sites to
getroutes (with layers ["auto.localchans","auto.sourcefree"]
and per-callsite maxparts), and adds a defensive has("old_state")
check in the channel-state-changed handler.

## Notes

- All three commits target CLN v26.06rc1 or later.  They use no
  API features predating askrene's first appearance in v24.08,
  so they are safe to apply on older CLN releases too.

- The Dowser uses a 1_000_000 sat probe with maxparts=10.  On
  signet, that's often too aggressive (many channels lack 1M
  sat capacity); askrene returns 205 errors with detailed
  messages like "isn't big enough to carry 1000000000msat" or
  "Total source capacity is only 420000000msat (in 2
  channels)".  Tuning probe_amount lower would let more
  small-channel candidates pass; consider revisiting if the
  rejection rate looks problematic in mainnet operation.

- **Dowser refactored to use getroutes**: Replaced legacy multi-call `getroute` + `listchannels` logic with a single `getroutes` probe (1,000,000 sat with maxparts=10), collapsing ~40 RPCs to 1. Removed self_id dependency and plumbing.

- **Matchmaker and ActiveProber migrated to getroutes**: Both subsystems now use `getroutes` with maxparts=1 instead of deprecated `getroute`. ActiveProber reconstructs sendpay-compatible hops by parsing typed fields from the getroutes response and removes the exclude parameter.

- **ChannelCreateDestroyMonitor hardened for omitted old_state**: Added defensive check for presence of `old_state` field in channel_state_changed notifications, defaulting to empty string when absent to prevent errors in CLN v26.06 where the "unknown" sentinel is removed.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ksedgwic/clboss/pull/313?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/Boss/Mod/ActiveProber.cpp
+++ b/Boss/Mod/ActiveProber.cpp
@@ -215,11 +215,16 @@ private:
 		});
 	}
 
+	/* First hop after `peer`: the peer's neighbor that we probe
+	 * towards.  Extracted into typed values from getroutes' path[0]
+	 * so we can rebuild the sendpay hop later without keeping the
+	 * raw Jsmn::Object around.
+	 */
+	Ln::NodeId id1;
 	Ln::Scid chan1;
+	std::uint32_t direction1;
 	Ln::Amount amount1;
 	std::uint32_t delay1;
-	/* Route except for the 0th hop from us to peer.  */
-	Jsmn::Object route;
 
 	Ev::Io<void> getroute() {
 		if (to_try.empty())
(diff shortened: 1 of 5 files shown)


ksedgwic:
Superseded by #314, #315, and #316

PR #316, opened 2026-05-20
cln-v26.06-compat: InvoicePayer pay -> xpay

ksedgwic:
## Summary

Stacked on top of #315  (`cln-v26.06-compat: FundsMover failure
feedback via askrene layer`).

`pay` was deprecated in CLN v26.06 (last supported v27.03,
tighter window than `getroute`'s v27.06).  Its replacement
`xpay` has been available since v24.11, so this is a straight
migration with no compatibility bridge required.

The sole producer of `Msg::PayInvoice` is `SwapManager` paying a
Boltz swap-out invoice (`Boss/Mod/SwapManager.cpp:541`), so this
code path is exclusively swap-out invoice payment -- real money
on prod.

- `pay` -> `xpay`, request field `bolt11` -> `invstring`.
- `maxfeepercent` (a percentage) replaced by `maxfee` (an
  absolute msat amount) computed from the decoded invoice
  amount.  The 0.5% fraction preserves the legacy non-MPP
  fee cap; xpay's own default of 1% would have doubled our
  historical fee budget, which matters for swap-out economics.
- Drop the MPP feature-bit detection (hex-parsing the bolt11
  features string).  It existed solely to pick between two
  `maxfeepercent` values for pre-xpay `pay`, whose pre-v0.10
  MPP fee-budget management required a generous 5% cap to
  succeed.  xpay's askrene-driven multi-path routing manages
  its own per-part budget.
- Drop the `Util::Str::HexParseFailure` catch path (no longer
  reachable) and the `Util/Str.hpp` include.

InvoicePayer reads no fields from the response (consumed as
`Jsmn::Object _`), so xpay's response-shape differences vs pay
are not visible at this caller.

CLN minimum unchanged from #314/#315 (xpay landed v24.11,
contemporaneous with askrene-create-layer/inform-channel).

- Migrates InvoicePayer from deprecated CLN `pay` RPC method to `xpay`, updating the request field from `bolt11` to `invstring` and computing an absolute `maxfee` (in msat) from the decoded invoice amount using a 0.5% cap instead of percentage-based `maxfeepercent`.
- Removes MPP feature-bit detection logic and associated hex-parsing error handling (`Util::Str::hexread`), since fee-cap selection is no longer needed with `xpay`.
- Test assertions updated to validate `xpay` request parameters (`invstring`, `retry_for: 1000`, and `maxfee: 5000`) instead of the prior `pay` parameters.

[...]

PR #315, opened 2026-05-20
cln-v26.06-compat: FundsMover failure feedback via askrene layer

ksedgwic:
## Summary

Stacked on top of #314  (`cln-v26.06-compat: migrate getroute
callers to getroutes`).

The fourth and last `getroute` call site in CLBOSS,
`FundsMover/Attempter.cpp`, was deliberately carved out of #314
because its exclude-vector pattern encoded failure feedback from
real payment attempts and needed a design step rather than a
mechanical port.

This PR migrates it to `getroutes` and replaces the local
excludes vector with a persistent askrene layer named
`"clboss"` (xpay's naming convention).  Onion-error feedback is
recorded into the layer via `askrene-inform-channel` (channel
capacity constraints) or `askrene-disable-node` (node-level
exclusions); the next `getroutes` call automatically routes
around the failure, both for this Attempter run and for every
subsequent CLBOSS attempt.

- `FundsMover/Main.cpp`: idempotent startup hook that
  `askrene-create-layer name="clboss" persistent=true`.
- `FundsMover/Attempter.cpp`: getroutes call with the layer
  applied; `maxfee_msat` set from the prorated fee budget so
  askrene refuses over-budget routes directly (replacing the
  legacy `fuzzpercent` ladder); typed `std::vector<Hop>`
  replaces the raw `Jsmn::Object route` member; on sendpay
  failure, parsed onion error feeds into
  `askrene-inform-channel` / `askrene-disable-node` and the
  call retries.

Field-name bridge from #314 is reused: path[] hop reads prefer
the new v26.06 names, fall back to the deprecated old names so
the code works on v26.04 prod and v26.06+ developer-mode signet
alike.

CLN minimum is v24.11 (when askrene-create-layer,
askrene-inform-channel, persistent layer support, and
short_channel_id_dir in path[] all landed together).

- Refactors FundsMover/Attempter to replace local excludes-vector pattern with persistent askrene layer (`clboss`). Routes now use `getroutes` with `maxfee_msat` set from prorated fee budget. On sendpay failures, parsed onion errors trigger `askrene-disable-node` (node-level) or `askrene-inform-channel` (channel capacity) feedback, enabling retry logic that routes around recorded failures.

[...]

PR #317, opened 2026-05-21
cln-v26.06-compat: ActiveProber: feed the clboss askrene layer

ksedgwic:
PR2 (#315) introduced a persistent askrene layer named "clboss"
to hold accumulated failure-feedback that future getroutes calls
automatically route around.  But the only writer to that layer
in PR2 is FundsMover/Attempter, which rarely fires real sendpay
attempts in practice -- on signet specifically, FundsMover's
prorated-fee-budget check gates out almost every attempt before
sendpay runs, so the layer in production stays empty.  The
architectural promise of "shared failure-feedback knowledge across
all CLBOSS subsystems" never materialises.

ActiveProber, by contrast, fires real sendpay probes against our
peers on a stochastic ~10-minute timer.  Each probe is a 2-hop
payment (us -> peer -> some neighbour of peer) with a bit-flipped
random payment_hash, designed to fail at the final hop with
WIRE_INCORRECT_OR_UNKNOWN_PAYMENT_DETAILS as the "success"
outcome.  This PR turns each probe outcome into one entry in the
clboss layer, making it the dominant feeder and finally giving
the layer real content.

* Probe outcome discrimination

For each probe (route = [chan0=us->peer, chan1=peer->id1]):

  failcode 0x400F (WIRE_INCORRECT_OR_UNKNOWN_PAYMENT_DETAILS) at
  the destination -> probe success -> askrene-inform-channel
  inform=succeeded chan1 amount1 (raises a lower-bound estimate
  on the channel's capacity).

  CHANNEL-level failure on chan1 -> askrene-inform-channel
  inform=constrained chan1 amount1 (upper-bound constraint).

  NODE-level failure (failcode bit 0x2000) -> askrene-disable-node
  on the erring node.

  Unparseable onion (code 202) -> askrene-disable-node on id1
  (best-guess culprit on a 2-hop route).

  echan == chan0 (our own outbound channel had the issue) -> skip.
  CLBOSS already has authoritative knowledge of local channel
  state from listpeerchannels; no need to re-record in the
  shared layer.

Because probe-success is the common outcome in healthy networks,
the success-feed becomes the dominant signal source -- not the
failure feed.  Over time the layer accumulates a steady stream
of "this channel pushed amount X recently" observations that
askrene treats as lower-bound capacity raises.

* Commits

  1. AskreneLayer: extract shared helpers for clboss layer writes.
     Pure refactor.  Moves the inform_channel / disable_node
     helpers and the "clboss" layer-name constant from
     Boss::Mod::FundsMover::Attempter (private members) into a
     new shared module Boss::Mod::AskreneLayer (free functions).
     FundsMover keeps calling the same logic, just via the new
     module.

  2. ActiveProber: feed sendpay failures into the clboss askrene
     layer.  Wires the catching<RpcError> on ActiveProber's
     waitsendpay to parse erring_index / erring_channel /
     erring_direction / erring_node / failcode out of the error
     data, discriminate per the table above, and call into
     AskreneLayer.

  3. ActiveProber: feed inform=succeeded on probe success.
     Symmetric positive feed -- when the probe's expected final-
     hop reject arrives, record the proven lower-bound capacity
     on chan1.

  4. tests/boss/test_askrene_layer: unit tests for the shared
     helpers.  Four mocked-RPC tests covering
     inform_channel_constrained, inform_channel_succeeded,
     disable_node, and the silent-RpcError-swallow contract
     (any layer-write call gracefully degrades to no-op if the
     RPC fails -- e.g. older CLN where askrene layers do not
     exist).

* Stack

Stacked on #316 (PR3, xpay migration).  Merge order:
#314 (PR1, getroutes) -> #315 (PR2, FundsMover layer) ->
#316 (PR3, xpay) -> this PR.

PR2's layer-write code paths (askrene-inform-channel,
askrene-disable-node) remain identical in behaviour after the
refactor in commit 1; the helpers are just hoisted into a
shared spot.  No regressions to FundsMover's existing flow.

- Wires ActiveProber to feed outcome feedback into the persistent "clboss" askrene layer by parsing waitsendpay errors and discriminating between probe successes (failcode 0x400F), channel-level failures, node-level failures (0x2000 bit set), and unparseable onions (code 202), with logic to skip writing for local outgoing channel failures.

- Extracts reusable AskreneLayer helpers (inform_channel_constrained, inform_channel_unconstrained, disable_node) and the clboss_layer_name constant into a new Boss::Mod::AskreneLayer module, with all helpers gracefully swallowing RpcErrors to degrade to no-op on older CLN versions.

- Refactors FundsMover/Attempter to use the centralized AskreneLayer helpers instead of local implementations (no behavioral change) and adds unit tests covering the new helpers' RPC request shapes and silent-error-swallow contract.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ksedgwic/clboss/pull/317?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/Boss/Mod/ActiveProber.cpp
+++ b/Boss/Mod/ActiveProber.cpp
@@ -1,4 +1,5 @@
 #include"Boss/Mod/ActiveProber.hpp"
+#include"Boss/Mod/AskreneLayer.hpp"
 #include"Boss/Mod/ChannelCandidateInvestigator/Main.hpp"
 #include"Boss/Mod/Rpc.hpp"
 #include"Boss/Msg/Init.hpp"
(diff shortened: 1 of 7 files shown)


ksedgwic:
Replacing w/ a better version shortly ...

PR #318, opened 2026-05-22
cln-v26.06-compat: FundsMover: clboss askrene layer maintenance — 24h aging, self-exclude, positive reinforcement

ksedgwic:
This PR bundles three small improvements to how FundsMover
  uses the persistent clboss askrene layer, all originally
  motivated by the PR1-5 migration to askrene-getroutes:

    1. (commit 1) FundsMover periodically calls askrene-age
       on the clboss layer with a 24h cutoff, so stale
       capacity constraints do not embed permanently into
       CLBOSS's routing model.

    2. (commit 1) FundsMover adds its own node_id to the
       clboss layer's disabled_nodes set at startup, so
       askrene-getroutes cannot return rebalance routes that
       loop back through us as a middle hop.  This restores
       the protection that the legacy getroute call provided
       via its exclude=[self_id] argument.

    3. (commit 2) FundsMover writes positive reinforcement
       (askrene-inform-channel inform=unconstrained) for each
       middle hop of a successful rebalance.  This balances
       the layer against PR2's failure-feedback writes, which
       are otherwise monotonically pessimistic.

  Together these three make the clboss layer self-correcting
  rather than monotonically darkening.  Each has direct
  observability via askrene-listlayers clboss and a new or
  existing FundsMover Debug log line.

### DIFF
--- a/Boss/Mod/AskreneLayer.cpp
+++ b/Boss/Mod/AskreneLayer.cpp
@@ -4,6 +4,7 @@
 #include"Jsmn/Object.hpp"
 #include"Json/Out.hpp"
 #include"Util/stringify.hpp"
+#include"Uuid.hpp"
 #include<assert.h>
 
 namespace Boss { namespace Mod { namespace AskreneLayer {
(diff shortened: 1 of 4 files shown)


ksedgwic:
Replacing w/ a better version shortly ...

